PR #189, opened 2017-08-04
fix README typo

kendase3:


kendase3:
you should submit a wording change; it could be Justin's Big Gondar PR

PR #192, opened 2017-08-07
update name to maker to be more consistent

kendase3:


kendase3:
I may have to change some jenkins goodies as well.

PR #195, opened 2017-08-08
Replace BOOL with a standard C++ bool

nicholasbishop:
Also replaced `TRUE` with `true` and `FALSE` with `false`.

I left a few instances of non-standard booleans alone so that I don't
accidentally break compatibility with any Windows API functions.


kendase3:
looks GOOD/good

PR #197, opened 2017-08-08
Clean up DeviceGuy

nicholasbishop:
Removed the two free functions and replace with constructors.


kendase3:
deviceguy really hummin' along now

PR #201, opened 2017-08-08
Update text shown to user without network

kendase3:
Addresses OVER-5205

### DIFF
--- a/src/image_select_page.cc
+++ b/src/image_select_page.cc
@@ -90,5 +90,5 @@ QUrl ImageSelectPage::getUrl() {
 
 void ImageSelectPage::handleNewestImageUrlError() {
   hasError = true;
-  wizard()->postError("An error has occurred fetching the latest image");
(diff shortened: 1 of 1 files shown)


nicholasbishop:
Is this change really accurate? There are a number of network conditions that might lead to this error, e.g. if we mess up permissions in AWS or there's an outage

kendase3:
I actually had the ticket wrong; it's OVER-5205.  I'd @ one of the concerned parties there; I have no preference.  We could make a ticket to try hitting an endpoint likely to be up like neverware.com (no joeks plz).  If we can't hit the known-good endpoint, we could display this message; else display a message saying 'our bad' or similar.

kendase3:
Imma push -f to update the "addresses".

nicholasbishop:
Commented in JIRA

kendase3:
i think this has been successfully 'why do we need this'd
closing for now

PR #203, opened 2017-08-09
Refactor to remove global variable selected_drive

nicholasbishop:


kendase3:
LG.  iirc the case wherein no drive is selected is prevented from happening by the device select page's validate page so that debug would never fire in the current state of things.

PR #206, opened 2017-08-10
Adjust device index by 130

kendase3:
DeviceGuy rework broke master for device index lookup.

### DIFF
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -232,7 +232,14 @@ bool DeviceSelectPage::validatePage() {
     return false;
   } else {
     unsigned int selected_index = selected->index;
-    wizard()->writeOperationPage.setDevice(drivelist.at(selected_index));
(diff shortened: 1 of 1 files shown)


nicholasbishop:
I'm probably misunderstanding something about how these various indices and numbers are supposed to work. Could you sketch it out for me?

kendase3:
The 130 is kind of explained by pbatard here https://github.com/neverware/gondar/blob/master/src/gondar.cc#L168

kendase3:
Actually, it looks like the min is 128 (0x80).  So this change will only make things work on my machine :(

nicholasbishop:
Oh! I see now that I misunderstood https://github.com/neverware/gondar/blob/b1ab2ddddb4876837cad46bda145be9e5e571e56/src/device.cc#L71

nicholasbishop:
I will put up a PR for that shortly

kendase3:
thx

PR #207, opened 2017-08-10
Fix device lookup

nicholasbishop:
In an earlier change I misinterpreted the selected index as being an
index into the container, rather than the device_num.


kendase3:
WoMM; LG

PR #212, opened 2017-08-18
logger: add support for DeviceGuy type

nicholasbishop:
This just makes logging a device a little more convenient.


kendase3:
LG once it looks good to the captain

PR #216, opened 2017-08-23
use dev endpoint for non-release, production for release builds

kendase3:


src/meepo.cc
```diff
@@ -31,9 +32,14 @@ const char path_auth[] = "/auth";
 const char path_sites[] = "/sites";
 const char path_downloads[] = "/downloads";
 
-// TODO(nicholasbishop): this is hardcoded to the staging URL for now
 QUrl createUrl(const QString& path) {
+#ifdef RELEASE
```
nicholasbishop: As with the `CHROMEOVER` build variable, we could add an `isRelease()` function to `util.h`.

nicholasbishop:
I was thinkin, since our usage of the meepo api is a pretty boring read-only operation, is using the grv meepo ever really necessary? Maybe we should just hard code the production URL.

kendase3:
I have an account on grv, but not the real one.  I like to use it for testing so I can have multiple sites associated with me, one site, etc.

kendase3:
I'd rather not mess around as an admin on the real meepo

kendase3:
I don't think this is adding much code complexity for that ability

nicholasbishop:
Sure, that sounds reasonable. Just wanted to be sure we needed it.

PR #217, opened 2017-08-28
Now send username/password as part of post data

kendase3:
Addresses OVER-5322
Addresses OVER-5317

### DIFF
--- a/src/meepo.cc
+++ b/src/meepo.cc
@@ -43,25 +43,19 @@ QUrl createUrl(const QString& path) {
   }
 }
 
-QString redactedUrl(QUrl url) {
-  QUrlQuery query(url);
-  url.setQuery(query);
-  return url.toString();
-}
-
 int siteIdFromUrl(const QUrl& url) {
   const auto path = url.path();
   const auto parts = path.split('/');
   const auto sites_index = parts.lastIndexOf("sites");
 
   if (sites_index == -1) {
-    LOG_ERROR << "failed to find 'sites' in " << redactedUrl(url);
+    LOG_ERROR << "failed to find 'sites' in " << url.toString();
     return -1;
   }
 
   const auto site_id_index = sites_index + 1;
   if (site_id_index >= parts.size()) {
-    LOG_ERROR << "url ended without a site ID: " << redactedUrl(url);
+    LOG_ERROR << "url ended without a site ID: " << url.toString();
     return -1;
   }
 
@@ -71,7 +65,7 @@ int siteIdFromUrl(const QUrl& url) {
   const auto site_id = site_id_str.toInt(&ok);
 
   if (!ok) {
-    LOG_ERROR << "site ID is not an integer: " << redactedUrl(url);
+    LOG_ERROR << "site ID is not an integer: " << url.toString();
     return -1;
   }
 
@@ -167,7 +161,7 @@ void Meepo::requestAuth(const QAuthenticator& auth) {
   QByteArray postData = credsJson.toUtf8();
   QByteArray postDataSize = QByteArray::number(postData.size());
   auto request = createAuthRequest(postDataSize);
-  LOG_INFO << "POST " << redactedUrl(request.url());
+  LOG_INFO << "POST " << request.url().toString();
   network_manager_.post(request, postData);
 }
 
@@ -185,7 +179,7 @@ void Meepo::handleAuthReply(QNetworkReply* reply) {
 
 void Meepo::requestSites() {
   const auto request = createSitesRequest(api_token_);
-  LOG_INFO << "GET " << redactedUrl(request.url());
+  LOG_INFO << "GET " << request.url().toString();
   network_manager_.get(request);
 }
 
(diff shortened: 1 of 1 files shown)


src/meepo.cc
```diff
@@ -45,8 +45,6 @@ QUrl createUrl(const QString& path) {
 
 QString redactedUrl(QUrl url) {
   QUrlQuery query(url);
-  query.removeQueryItem("password");
-  query.removeQueryItem("token");
   url.setQuery(query);
   return url.toString();
 }
```
nicholasbishop: This function can be removed, doesn't do anything now

kendase3:
I didn't give much thought to string concat method etc.  I made sure that the user/pass are not logged to terminal.

src/meepo.cc
```diff
@@ -80,16 +78,13 @@ int siteIdFromUrl(const QUrl& url) {
   return site_id;
 }
 
-QNetworkRequest createAuthRequest(const QAuthenticator& auth) {
+QNetworkRequest createAuthRequest(QByteArray postDataSize) {
```
nicholasbishop: `const QByteArray&`

src/meepo.cc
```diff
@@ -80,16 +78,13 @@ int siteIdFromUrl(const QUrl& url) {
...
   auto url = createUrl(path_auth);
-  QUrlQuery query;
-  query.addQueryItem("email", auth.user());
-  query.addQueryItem("password", auth.password());
-  url.setQuery(query);
-
   QNetworkRequest request(url);
   request.setHeader(QNetworkRequest::ContentTypeHeader,
                     "application/x-www-form-urlencoded");
+  request.setRawHeader("Content-Type", "application/json");
```
nicholasbishop: Looks like you are overwriting the content type header here, should just change the earlier header (and no need to use raw)

src/meepo.cc
```diff
@@ -80,16 +78,13 @@ int siteIdFromUrl(const QUrl& url) {
...
+  request.setRawHeader("Content-Length", postDataSize);
```
nicholasbishop: No need to set this explicitly, QNetworkAccessManager::post already does this

src/meepo.cc
```diff
@@ -163,9 +158,17 @@ Meepo::Sites Meepo::sites() const {
 }
 
 void Meepo::requestAuth(const QAuthenticator& auth) {
-  const auto request = createAuthRequest(auth);
+  // TODO: sanitize?
+  QString credsJson = "{\"email\":\""
```
justinoue: is there some sort of json library to do this? something to handle auto escaping the user input and what not.

src/meepo.cc
```diff
@@ -163,9 +152,17 @@ Meepo::Sites Meepo::sites() const {
 }
 
 void Meepo::requestAuth(const QAuthenticator& auth) {
-  const auto request = createAuthRequest(auth);
-  LOG_INFO << "POST " << redactedUrl(request.url());
-  network_manager_.post(request, QByteArray());
+  // TODO: sanitize?
+  QString credsJson = "{\"email\":\""
```
justinoue: is there some sort of json library to do this? something to handle auto escaping the user input and what not.

nicholasbishop: Yep, this should use the Qt JSON library

justinoue:
looks good to me as long as it works

src/meepo.cc
```diff
@@ -163,9 +149,14 @@ Meepo::Sites Meepo::sites() const {
...
 
 void Meepo::requestAuth(const QAuthenticator& auth) {
-  const auto request = createAuthRequest(auth);
-  LOG_INFO << "POST " << redactedUrl(request.url());
-  network_manager_.post(request, QByteArray());
+  QJsonObject json;
+  json["email"] = auth.user();
+  json["password"] = auth.password();
+  QJsonDocument doc(json);
+  QString strJson(doc.toJson(QJsonDocument::Compact));
```
nicholasbishop: There's some unnecessary conversion here; the return type of `toJson` is already a `QByteArray`: http://doc.qt.io/qt-5/qjsondocument.html#toJson

So that value can go directly to the `post` call.

kendase3: so you're suggesting i can remove the toUtf8() ?

nicholasbishop: Not just the `toUtf8`, you can also remove the conversion to a `QString`. E.g.

```c++
const QJsonDocument doc(json);
network_manager_.post(request, doc.toJson(QJsonDocument::Compact));
```

Otherwise you're converting from `QByteArray` -> `QString` -> `QByteArray`.

kendase3: and this compiles for you?

nicholasbishop: Yep, I just tried it under both Lin and Win. Here's the full function as I compiled it:

```c++
void Meepo::requestAuth(const QAuthenticator& auth) {
  QJsonObject json;
  json["email"] = auth.user();
  json["password"] = auth.password();
  const QJsonDocument doc(json);
  auto request = createAuthRequest();
  LOG_INFO << "POST " << request.url().toString();
  network_manager_.post(request, doc.toJson(QJsonDocument::Compact));
}
```

kendase3: should be good now.  i had tried removing the toString() calls to some urls and merged with the latest PR of yours but got some crazy <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<< fullscreen error output.  i just abandoned that merge for now.

PR #218, opened 2017-08-28
logging: specialize QUrl for easier logging

nicholasbishop:
This allows logging of a `QUrl` without calling `toString()`.


nicholasbishop:
(Incidentally, I didn't simplify the logs in `meepo.cc` since it'd conflict with the username+password PR)

PR #220, opened 2017-09-05
Chromeover flow now shows build from all products

kendase3:
Addresses OVER-5337

Distributable images are excluded.  Previous behavior worked well for CloudReady sites, but would not display other products, and would crash if a customer only had VDI or 365 products available.

### DIFF
--- a/src/gondarimage.h
+++ b/src/gondarimage.h
@@ -20,11 +20,10 @@ class GondarImage {
   public:
     GondarImage() {
     }
-    GondarImage(QString productIn, QString imageNameIn, QUrl urlIn) {
-        this->product = productIn;
-        this->imageName = imageNameIn;
-        this->url = urlIn;
-    }
+    GondarImage(QString productIn, QString imageNameIn, QUrl urlIn) : 
+        product(productIn), 
+        imageName(imageNameIn),
+        url(urlIn) { }
     QString getCompositeName() {
         return product + " " + imageName;
     }
--- a/src/gondarimage.h
+++ b/src/gondarimage.h
@@ -1,3 +1,17 @@
+// Copyright 2017 Neverware
+//
+// This program is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
 
 #ifndef GONDARIMAGE_H
 #define GONDARIMAGE_H
--- a/src/gondarimage.h
+++ b/src/gondarimage.h
@@ -1,0 +1,31 @@
+
+#ifndef GONDARIMAGE_H
+#define GONDARIMAGE_H
+
+class GondarImage {
+  public:
+    GondarImage() {
+        this->product = QString("UHOH");
+        this->imageName = QString("OHDEAR");
+        this->url = QUrl();
+    }
+    GondarImage(QString productIn, QString imageNameIn, QUrl urlIn) {
+        this->product = productIn;
+        this->imageName = imageNameIn;
+        this->url = urlIn;
+    }
+    QString getCompositeName() {
+        return product + " " + imageName;
+    }
+    bool is32Bit() {
+        return imageName.contains("32-bit", Qt::CaseInsensitive);
+    }
+    bool isDeployable() {
+        return imageName.contains("deployable", Qt::CaseInsensitive);
+    }
+    QString product;
+    QString imageName;
+    QUrl url;
+};
+
+#endif
--- a/src/gondarimage.h
+++ b/src/gondarimage.h
@@ -17,25 +17,18 @@
 #define GONDARIMAGE_H
 
 class GondarImage {
-  public:
-    GondarImage() {
-    }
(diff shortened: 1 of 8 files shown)


src/gondarimage.h
```diff
@@ -0,0 +1,31 @@
+
```
nicholasbishop: Copyright header

src/gondarimage.h
```diff
@@ -0,0 +1,31 @@
...
+#ifndef GONDARIMAGE_H
+#define GONDARIMAGE_H
+
+class GondarImage {
+  public:
+    GondarImage() {
+        this->product = QString("UHOH");
+        this->imageName = QString("OHDEAR");
```
nicholasbishop: What're these magic strings?

src/gondarimage.h
```diff
@@ -0,0 +1,31 @@
...
+        this->url = QUrl();
```
nicholasbishop: This isn't necessary, the default constructor is already run

src/gondarimage.h
```diff
@@ -0,0 +1,31 @@
...
+    }
+    GondarImage(QString productIn, QString imageNameIn, QUrl urlIn) {
```
nicholasbishop: Prefer the member initialization syntax:

    : product(productIn), blah(blahIn), ... {}

src/gondarimage.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef GONDARIMAGE_H
+#define GONDARIMAGE_H
+
+class GondarImage {
+ public:
+  GondarImage() {}
+  GondarImage(QString productIn, QString imageNameIn, QUrl urlIn)
+      : product(productIn), imageName(imageNameIn), url(urlIn) {}
+  QString getCompositeName() { return product + " " + imageName; }
+  bool is32Bit() { return imageName.contains("32-bit", Qt::CaseInsensitive); }
```
nicholasbishop: bool is32Bit() const

src/gondarimage.h
```diff
@@ -0,0 +1,34 @@
...
+  bool isDeployable() {
```
nicholasbishop: bool isDeployable() const

src/image_select_page.cc
```diff
@@ -68,23 +84,55 @@ int ImageSelectPage::nextId() const {
...
+  if (image.is32Bit()) {
+    layout.addWidget(&thirtyTwoDetails);
+  }
 }
 
-void ImageSelectPage::set64Url(QUrl url_in) {
-  newestImageUrl.set64Url(url_in);
+void ImageSelectPage::addImages(QList<GondarImage> images) {
+  GondarImage curImage;
+  foreach (curImage, images) {
```
nicholasbishop: Prefer the C++11 syntax:

    for (const auto& curImage : images) {

src/image_select_page.cc
```diff
@@ -68,23 +84,55 @@ int ImageSelectPage::nextId() const {
...
 }
 
-void ImageSelectPage::set32Url(QUrl url_in) {
-  newestImageUrl.set32Url(url_in);
+void ImageSelectPage::addImage(GondarImage image) {
+  // do not list deployable images
+  if (image.isDeployable()) {
+    return;
+  }
+  // FIXME: leak
```
nicholasbishop: I don't think this is actually a leak, when you place a widget inside a parent the parent takes ownership and will `delete` the child

src/meepo.cc
```diff
@@ -207,22 +206,22 @@ void Meepo::handleDownloadsReply(QNetworkReply* reply) {
...
-    QJsonObject download = downloadsArray.at(i).toObject();
-    if (download["title"] == "64-Bit") {
-      QUrl url64(download["url"].toString());
-      site->set64Url(url64);
-    } else if (download["title"] == "32-Bit") {
-      QUrl url32(download["url"].toString());
-      site->set32Url(url32);
+  QJsonObject productsObj = jsonObj["links"].toObject();
+  QJsonDocument doc(productsObj);
+  QString productsStr(doc.toJson(QJsonDocument::Compact));
```
nicholasbishop: This variable appears to be unused

src/gondarsite.h
```diff
@@ -19,19 +19,20 @@
...
  public:
   GondarSite(int siteIdIn, QString siteNameIn);
   int getSiteId() const;
   const QString& getSiteName() const;
-  void set32Url(QUrl url);
-  void set64Url(QUrl url);
-  QUrl get32Url() const;
-  QUrl get64Url() const;
+  QList<GondarImage> getImages() const;
+  void addImage(QString product, QString imageName, QUrl url);
```
nicholasbishop: Nitpick: change to `void addImage(const GondarImage& image)`, that way if we change `GondarImage` we don't have to update this interface

src/meepo.cc
```diff
@@ -207,22 +206,20 @@ void Meepo::handleDownloadsReply(QNetworkReply* reply) {
...
-    if (download["title"] == "64-Bit") {
-      QUrl url64(download["url"].toString());
-      site->set64Url(url64);
-    } else if (download["title"] == "32-Bit") {
-      QUrl url32(download["url"].toString());
-      site->set32Url(url32);
+  QJsonObject productsObj = jsonObj["links"].toObject();
+
+  int productItr = 0;
+  for (QJsonValueRef product : productsObj) {
```
nicholasbishop: This can be simplified a bit more, something like this:

```c++
const auto productsObj = jsonObj["links"].toObject();

  for (const auto& product : productsObj.keys()) {
    for (const auto& image : productsObj[product].toArray()) {
      const auto imageObj = image.toObject();
      const auto imageName(imageObj["title"].toString());
      const QUrl url(imageObj["url"].toString());

      LOG_INFO << product << ", " << imageName << ", " << url.toString();
      site->addImage(product, imageName, url);
    }
  }
```

PR #221, opened 2017-09-06
Added 'forgot password?' link

kendase3:
Addresses OVER-5309

### DIFF
--- a/src/chromeover_login_page.cc
+++ b/src/chromeover_login_page.cc
@@ -35,17 +35,17 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
   meanWordsLabel.setText("That's not right!  Try again.");
   QString forgotText = "<a href=\"https://my." + gondar::getDomain() +
                        "/forgot-password\">Forgot your password?</a>";
-  frogotLabel.setText(forgotText);
-  frogotLabel.setTextFormat(Qt::RichText);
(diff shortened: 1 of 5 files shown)


src/chromeover_login_page.cc
```diff
@@ -32,11 +32,19 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
   passwordLineEditLabel.setBuddy(&passwordLineEdit);
   passwordLineEditLabel.setText("Password:");
   meanWordsLabel.setText("That's not right!  Try again.");
+  frogotLabel.setText(
+      "<a href=\"https://my.neverware.com/forgot-password\">Forgot your "
```
nicholasbishop: The domain here should somehow be synced with the one in https://github.com/neverware/gondar/blob/82948eca6e2baad1c2b5861293e2e9dcf2c85966/src/meepo.cc#L36

src/chromeover_login_page.h
```diff
@@ -46,7 +46,7 @@ class ChromeoverLoginPage : public gondar::WizardPage {
   QLineEdit passwordLineEdit;
   QLabel passwordLineEditLabel;
   QLabel meanWordsLabel;
-  QLabel frogotLabel;
+  QLabel forgotLabel;
```
nicholasbishop: gondar used 2 b about the 🎵 rip 2017

PR #222, opened 2017-09-06
Kendall version metric 01

kendase3:


src/metric.cc
```diff
@@ -70,6 +70,14 @@ QByteArray getMetricsApiKey() {
...
 #endif
 }
+
+QString getGondarVersion() {
+#ifdef GONDAR_VERSION
+  return QString(GONDAR_VERSION);
+#else
+  return QString();
+#endif
+}
```
nicholasbishop: I'd mildly prefer reading the version number from a file in the repo, rather than passing it in from the build system. Kinda like what we do in chromeover-build: https://github.com/neverware/chromeover-build/blob/neverware-d59/version_string

Maybe just have this function directly return a version string? In master it'd be `QString()`, and the release branch can return "1.9" or whatever.

PR #223, opened 2017-09-06
more appropriate logging on zip complete

kendase3:


src/unzipthread.cc
```diff
@@ -36,5 +36,5 @@ void UnzipThread::run() {
     LOG_ERROR << "unzip failed: " << exc.what();
     filename = QString();
   }
-  qDebug() << "worker thread says complete";
+  LOG_INFO << "Unzip succeeded";
```
nicholasbishop: This isn't accurate; either it should be changed to something like "Unzip operation ended", or it should be moved inside the `try` block so that it doesn't say "Unzip succeeded" after an error.

PR #225, opened 2017-09-07
Check if there is a newer version, display message

kendase3:
Addresses OVER-5336

Also adds the current version number to the window title

### DIFF
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -148,7 +148,7 @@ qint64 GondarWizard::getRunTime() {
   return p_->runTime.secsTo(QDateTime::currentDateTime());
 }
 
-void GondarWizard::showUpdateNeeded(QString latestVersionString) {
+void GondarWizard::showUpdateNeeded(const QString& latestVersionString) {
   QString baseUrl = "https://usb-maker-downloads.neverware.com/stable/";
   QString product;
   QString filename = "cloudready-usb-maker.exe";
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -149,12 +149,23 @@ qint64 GondarWizard::getRunTime() {
 }
 
 void GondarWizard::showUpdateNeeded(QString latestVersionString) {
+  QString downloadUrl;
+  if (gondar::isChromeover()) {
+    downloadUrl =
+        "https://d20slnb0el8xd1.cloudfront.net/stable/cloudready/"
+        "cloudready-usb-maker.exe";
+  } else {
+    downloadUrl =
+        "https://d20slnb0el8xd1.cloudfront.net/stable/cloudready-free/"
+        "cloudready-usb-maker.exe";
+  }
   QMessageBox::information(this, "CloudReady USB Maker",
                            "<a "
-                           "href=\"https://d2znie75my0pef.cloudfront.net/20/"
-                           "cloudready-free/cloudready-usb-maker.exe\">A new "
-                           "version "
-                           "of CloudReady USB Maker (" +
+                           "href=\"" +
+                               downloadUrl +
+                               "\">A new "
+                               "version "
+                               "of CloudReady USB Maker (" +
                                latestVersionString + ") is available.</a>");
 }
 
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -62,7 +62,7 @@ GondarWizard::GondarWizard(QWidget* parent)
   setPage(Page_writeOperation, &writeOperationPage);
   setPage(Page_error, &p_->errorPage);
   setWizardStyle(QWizard::ModernStyle);
-  setWindowTitle("CloudReady USB Maker " + gondar::getGondarVersion());
+  setWindowTitle(tr("CloudReady USB Maker"));
   setPixmap(QWizard::LogoPixmap, QPixmap(":/images/crlogo.png"));
 
   setButtonText(QWizard::CustomButton1, "Make Another USB");
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -15,6 +15,7 @@
 
 #include "gondarwizard.h"
 
+#include <QMessageBox>
 #include <QTimer>
 
 #include "about_dialog.h"
@@ -25,6 +26,7 @@
 #include "log.h"
 #include "metric.h"
 #include "site_select_page.h"
+#include "util.h"
 
 class GondarWizard::Private {
  public:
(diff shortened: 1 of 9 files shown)


src/gondarwizard.cc
```diff
@@ -143,3 +147,43 @@ void GondarWizard::catchError(const QString& error) {
 qint64 GondarWizard::getRunTime() {
   return p_->runTime.secsTo(QDateTime::currentDateTime());
 }
+
+void GondarWizard::showUpdateNeeded(QString latestVersionString) {
+  LOG_WARNING << "show update needed fires";
+  QMessageBox::information(
+      this, "CloudReady USB Maker",
+      "<a href=\"https://guide.neverware.com/supported-devices\">A new version "
```
nicholasbishop: wrong link?

kendase3: yeah that's intentional; i'll add the TODO

kendase3: i've just added the url to the .exe for now.  there's no webpage for gondar yet.

src/gondarwizard.cc
```diff
@@ -60,7 +62,7 @@ GondarWizard::GondarWizard(QWidget* parent)
   setPage(Page_writeOperation, &writeOperationPage);
   setPage(Page_error, &p_->errorPage);
   setWizardStyle(QWizard::ModernStyle);
-  setWindowTitle(tr("CloudReady USB Maker"));
+  setWindowTitle("CloudReady USB Maker " + gondar::getGondarVersion());
```
nicholasbishop: I don't think the version should be in the main application titlebar, it's not common to do that. It would make sense to show in the text of the about dialog though.

src/gondarwizard.cc
```diff
@@ -143,3 +147,55 @@ void GondarWizard::catchError(const QString& error) {
...
+  QString downloadUrl;
+  if (gondar::isChromeover()) {
+    downloadUrl =
+        "https://usb-maker-downloads.neverware.com/stable/cloudready/"
+        "cloudready-usb-maker.exe";
+  } else {
+    downloadUrl =
+        "https://usb-maker-downloads.neverware.com/stable/cloudready-free/"
+        "cloudready-usb-maker.exe";
+  }
```
nicholasbishop: Should factor out the common bit, `const baseUrl QUrl("https://usb-maker-downloads.neverware.com/stable")` or similar

src/gondarwizard.cc
```diff
@@ -143,3 +147,55 @@ void GondarWizard::catchError(const QString& error) {
 qint64 GondarWizard::getRunTime() {
   return p_->runTime.secsTo(QDateTime::currentDateTime());
 }
+
+void GondarWizard::showUpdateNeeded(QString latestVersionString) {
```
nicholasbishop: `const QString&`

src/gondarwizard.cc
```diff
@@ -143,3 +147,55 @@ void GondarWizard::catchError(const QString& error) {
...
+
+void GondarWizard::handleVersionReply(QNetworkReply* reply) {
+  QString latestVersionString = reply->readAll();
+  double latestVersion = latestVersionString.toDouble();
+  double currentVersion = gondar::getGondarVersion().toDouble();
+  LOG_INFO << "Latest version: " << latestVersion
+           << ", currentVersion: " << currentVersion;
+  if (currentVersion < latestVersion) {
+    showUpdateNeeded(latestVersionString);
+  }
```
nicholasbishop: In the interest of preventing the `GondarWizard` class from being a kitchen sink, how about a new class in its own files to handle getting/checking the version and creating the "pls download now" dialog. `gondar::UpdateCheck` or something like that.

src/gondarwizard.cc
```diff
@@ -143,3 +147,55 @@ void GondarWizard::catchError(const QString& error) {
...
+        "cloudready-usb-maker.exe";
+  }
+  QMessageBox::information(this, "CloudReady USB Maker",
+                           "<a "
+                           "href=\"" +
+                               downloadUrl +
+                               "\">A new "
+                               "version "
+                               "of CloudReady USB Maker (" +
+                               latestVersionString + ") is available.</a>");
```
nicholasbishop: Suggestion for the message box:

```c++
QMessageBox box(
      QMessageBox::Information, tr("Update Available"),
      tr("A new version of CloudReady USB Maker is now available."));
  box.setInformativeText(tr(
      "Click “Download Update” to download version %2 in your web browser. "
      "When your download starts you can close CloudReady USB Maker. "
      "To complete the upgrade just launch the new version.")
      .arg(latestVersionString));
  auto* reject =
    box.addButton(tr("&Not Now"), QMessageBox::RejectRole);
  auto* accept =
    box.addButton(tr("&Download Update"), QMessageBox::AcceptRole);
  box.setEscapeButton(reject);
  box.setDefaultButton(accept);
  box.exec();

  if (box.clickedButton() == accept) {
    QDesktopServices::openUrl(downloadUrl);
  }
```

The text changes are to help the user understand what kind of upgrade this is (i.e. a very simple manual process) and to use clearer verbs in the buttons ("Download Update" instead of "OK").

src/gondarwizard.cc
```diff
@@ -143,3 +147,55 @@ void GondarWizard::catchError(const QString& error) {
...
+  QNetworkAccessManager* networkManager = getNetworkManager();
+  connect(networkManager, &QNetworkAccessManager::finished, this,
+          &GondarWizard::handleVersionReply);
+  if (gondar::getGondarVersion().isEmpty()) {
+    LOG_INFO << "Skipping latest version check for dev build";
+    return;
+  }
+  networkManager->get(QNetworkRequest(
+      // TODO: create actual endpoint file
+      QUrl("http://neverware.com/hypothetical-latest-gondar-release-file")));
```
nicholasbishop: Filed https://neverware.atlassian.net/browse/OVER-5376 for this. In that bug I request a JSON file. It's obviously not strictly needed in this case, but it's a nice hedge in case we want to add some extra data without breaking older clients.

src/gondarwizard.cc
kendase3: I put it in the about dialog, above the html content.  it only shows when the version is populated

src/gondarwizard.cc
kendase3: i can make a wizardbuddy for this as requested

nicholasbishop:
FYI since you are OOO at the moment I pushed a couple commits to merge in latest master and `make format`.

src/gondarwizard.cc
kendase3: i made it

kendase3:
FYI since you are ZZZ at the moment I pushed a couple commits

PR #226, opened 2017-09-08
Create release script for Gondar

kendase3:
Addresses OVER-5333

### DIFF
--- a/gondar-release.py
+++ b/gondar-release.py
@@ -26,10 +26,11 @@ WORKDIR = 'gondar-release-workdir'
 SOURCE_BUCKET = 'neverware-gondar-release-candidates'
 SOURCE_FILENAME = 'cloudready-usb-maker.exe'
 DEST_BUCKET = 'neverware-cloudready-usb-releases'
+
 BEEROVER_UPLOAD_NAME = 'cloudready-free'
 CHROMEOVER_UPLOAD_NAME = 'cloudready'
 
-DEST_KEY_PREFIX = 'stable'
+STABLE_KEY_PREFIX = 'stable'
 DEST_FILENAME = 'cloudready-usb-maker.exe'
 
 def clean_workdir(makedir=True):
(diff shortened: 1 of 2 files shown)


gondar-release.py
```diff
@@ -0,0 +1,105 @@
...
+  prod_client.upload_file(local_path, DEST_BUCKET, dest_key_path)
+  # make the file public
+  prod_client.put_object_acl(ACL='public-read',
+                             Bucket=DEST_BUCKET,
+                             Key=dest_key_path)
+
+def parse_args():
+  parser = argparse.ArgumentParser()
+  parser.add_argument('candidate')
+  parser.add_argument('--beerover', action='store_true')
```
nicholasbishop: The way #225 is currently written there is a single version check used for both beerover and chromeover. That suggests that we always have to release beerover and chromeover in sync, otherwise some users will get "upgrade now!" messages despite no new version being available.

So does it make sense to have separate chromeover/beerover release modes here? Or should it just do both at the same time?

kendase3: that's something i admittedly didn't think about.  i think it would save us a lot of headaches to just always release the two together with the same version number.  the worst degradation i can think of is that users would be nagged to upgrade to a version with trivial or no new features for their version.

kendase3:
take another gondar gander when you guys get a chance.  we now release to the appropriate version number folder in addition to updating stable

gondar-release.py
```diff
@@ -0,0 +1,122 @@
...
+  print("archiving release to key={}".format(candidate_key_path))
+  prod_client.upload_file(local_path, DEST_BUCKET, stable_key_path)
+  prod_client.upload_file(local_path, DEST_BUCKET, candidate_key_path)
+  # make the files public
+  set_public_read(prod_client, DEST_BUCKET, stable_key_path)
+  set_public_read(prod_client, DEST_BUCKET, candidate_key_path)
+
+def parse_args():
+  parser = argparse.ArgumentParser()
+  parser.add_argument('candidate')
```
nicholasbishop: This could use a `help=`, not obvious whether it wants a local path or an "s3://" path or whatever

gondar-release.py
```diff
@@ -0,0 +1,122 @@
...
+  candidate_key_path = '/'.join([candidate,
+                                 product_upload_name,
+                                 DEST_FILENAME])
+  print("uploading to key={}".format(stable_key_path))
+  print("archiving release to key={}".format(candidate_key_path))
+  prod_client.upload_file(local_path, DEST_BUCKET, stable_key_path)
+  prod_client.upload_file(local_path, DEST_BUCKET, candidate_key_path)
+  # make the files public
+  set_public_read(prod_client, DEST_BUCKET, stable_key_path)
+  set_public_read(prod_client, DEST_BUCKET, candidate_key_path)
```
nicholasbishop: Nitpick, could move this into a loop:

```python
for key in (candidate_key_path, stable_key_path):
    prod_client.upload_file(local_path, DEST_BUCKET, key)
    set_public_read(prod_client, DEST_BUCKET, key)
```

gondar-release.py
```diff
@@ -0,0 +1,122 @@
...
+  # then we upload it to the appropriate prod bucket and set public readable
+  upload_file(prod_client, local_path, candidate, beerover)
+
+def main():
+  args = parse_args()
+  dev_client, prod_client = get_clients()
+  # release beerover
+  release(dev_client, prod_client, args.candidate, True)
+  # release chromeover
+  release(dev_client, prod_client, args.candidate, False)
```
nicholasbishop: Nitpick, boolean arguments aren't very readable. Could change it to

    release(dev_client, prod_client, args.candidate, beerover=True)

kendase3:
@dgs3 any more input on this or is it good to merge?

gondar-release.py
```diff
@@ -0,0 +1,121 @@
...
+                        Key=key)
+
+def upload_file(prod_client, local_path, candidate, beerover):
+  # then we upload it to the prod bucket
+  if beerover:
+    product_upload_name = BEEROVER_UPLOAD_NAME
+  else:
+    product_upload_name = CHROMEOVER_UPLOAD_NAME
+  stable_key_path = '/'.join([STABLE_KEY_PREFIX,
+                            product_upload_name,
```
dgs3: nitpick: indentation.

PR #229, opened 2017-09-12
Now use a JSON file in prod AWS for gondar update check

kendase3:
Addresses OVER-5376

This should work already; just change the current version in util.cc; 1.8 and below should pop up the update

### DIFF
--- a/src/update_check.cc
+++ b/src/update_check.cc
@@ -44,12 +44,13 @@ void UpdateCheck::start(QWidget* parent) {
           &UpdateCheck::handleReply);
 
   manager.get(QNetworkRequest(
-      // TODO: create actual endpoint file
-      QUrl("http://neverware.com/hypothetical-latest-gondar-release-file")));
+      QUrl("https://usb-maker-downloads.neverware.com/gondar-newest.json")));
 }
 
 void UpdateCheck::handleReply(QNetworkReply* reply) {
-  const QString latestVersionString = reply->readAll();
+  //TODO: parse json
+  const QString latestVersionString =
+      gondar::jsonFromReply(reply)["newest_release"].toString();
   const double latestVersion = latestVersionString.toDouble();
   const double currentVersion = gondar::getGondarVersion().toDouble();
   LOG_INFO << "Latest version: " << latestVersion
--- a/src/update_check.cc
+++ b/src/update_check.cc
@@ -44,7 +44,7 @@ void UpdateCheck::start(QWidget* parent) {
           &UpdateCheck::handleReply);
 
   manager.get(QNetworkRequest(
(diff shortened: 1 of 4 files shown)


nicholasbishop:
What do you think about adding an API key to the S3 path as suggested in https://neverware.atlassian.net/browse/OVER-5376?

kendase3:
sure we can do that; i already had dave make this path but i'm sure he can update it.  i thought the intent was mostly so we could expand it to contain more information or something.  i'm not sure i entirely understand the use case for removing this key/value pair, but i can do that.

kendase3:
i don't see the functional difference is between what you suggested and moving from https://usb-maker-downloads.neverware.com/gondar-newest.json to https://usb-maker-downloads.neverware.com/metadata.json once the file contains more than the newest version

nicholasbishop:
Possibly we are talking past each other, but my main point was that the URL that Gondar gets this info from should, in some form, include a "v1" in the path so that if we have to making a breaking change to the data (as opposed to merely adding fields) we can do so without messing up existing Gondar clients.

kendase3:
so what outstanding work is there on this?

src/update_check.cc
```diff
@@ -44,12 +44,13 @@ void UpdateCheck::start(QWidget* parent) {
...
 
   manager.get(QNetworkRequest(
-      // TODO: create actual endpoint file
-      QUrl("http://neverware.com/hypothetical-latest-gondar-release-file")));
+      QUrl("https://usb-maker-downloads.neverware.com/api/v1/metadata.json")));
 }
 
 void UpdateCheck::handleReply(QNetworkReply* reply) {
-  const QString latestVersionString = reply->readAll();
+  //TODO: parse json
```
nicholasbishop: Out of date todo

src/update_check.cc
```diff
@@ -44,12 +44,13 @@ void UpdateCheck::start(QWidget* parent) {
           &UpdateCheck::handleReply);
 
   manager.get(QNetworkRequest(
-      // TODO: create actual endpoint file
-      QUrl("http://neverware.com/hypothetical-latest-gondar-release-file")));
+      QUrl("https://usb-maker-downloads.neverware.com/api/v1/metadata.json")));
```
nicholasbishop: Currently this URL doesn't exist

kendase3: yeah, dave'll make it on the prod aws.  he made the other one last night.  i can make a ticket for it.  i don't think it should be a blocker for this PR; the 'hypothetical-latest-gondar-release-file' doesn't exist either.

kendase3: https://neverware.atlassian.net/browse/OVER-5394

PR #230, opened 2017-09-18
Gray out disks that are too small

kendase3:
Addresses OVER-5402

### DIFF
--- a/src/device.h
+++ b/src/device.h
@@ -27,7 +27,9 @@ A linkedlist of device info passed between the C and C++ layer
 class DeviceGuy {
  public:
   DeviceGuy(const DeviceGuy& other) = default;
-  DeviceGuy(uint32_t device_num, const std::string name, bool big = true);
+  DeviceGuy(uint32_t device_num,
+            const std::string name,
+            uint64_t num_bytes = 0);
 
   bool operator==(const DeviceGuy& other) const;
 
@@ -35,8 +37,7 @@ class DeviceGuy {
 
   uint32_t device_num = 0;
   std::string name;
-  // whether or not device is sufficiently large
-  bool big;
+  uint64_t num_bytes;
 };
 
 typedef std::vector<DeviceGuy> DeviceGuyList;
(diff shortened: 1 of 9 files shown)


src/device.h
```diff
@@ -27,14 +27,16 @@ A linkedlist of device info passed between the C and C++ layer
...
+  DeviceGuy(uint32_t device_num, const std::string name, bool big = true);
 
   bool operator==(const DeviceGuy& other) const;
 
   std::string toString() const;
 
   uint32_t device_num = 0;
   std::string name;
+  // whether or not device is sufficiently large
+  bool big;
```
nicholasbishop: I think it would be preferable to pull the actual size into the Device structure and then let the UI layer decide what is big enough.
I'd suggest something like:

```c++
uint64_t capacity_in_bytes_ = 0;
```

src/device.h
```diff
@@ -27,16 +27,17 @@ A linkedlist of device info passed between the C and C++ layer
...
 
   bool operator==(const DeviceGuy& other) const;
 
   std::string toString() const;
 
   uint32_t device_num = 0;
   std::string name;
-  // whether or not device is sufficiently large
-  bool big;
+  uint64_t num_bytes;
```
nicholasbishop: I'd recommend to set this to ` = 0` here rather than using the default parameter. Primitive values aren't autoinitialized in C++, so putting an inline default in the declaration ensures the value is always well defined.

kendase3:
It seems like we would want a default argument of some sane disk size so that tests can by default make valid disks.  when i had the 'big' abstraction, big was true by default.

nicholasbishop:
For the test cases, how about just adding a constant set to ~16GiB and pass that parameter to each disk?

kendase3:
what about src/stubs.cc

nicholasbishop:
Wouldn't that make the behavior of the tests change when run on Linux vs. Windows?

kendase3:
sorry, what's the "that" in "wouldn't that..."

kendase3:
i wasn't proposing a solution, i meant there's another set of initializations that we'd presumably want non-zero constructors for in src/stubs.cc:
devices.emplace_back(DeviceGuy(0, "stubdevice0"));

nicholasbishop:
Oh, I misunderstood. I agree that stubs.cc should have a fake size that isn't zero. Perhaps add another device to the stubs list? Then we could have a couple "big enough" drives and one "too small" drive.

PR #231, opened 2017-09-18
about dialog: use readUtf8File

nicholasbishop:
This relates to Kendall's work on better integration tests. They are
failing with an exception that didn't give enough information. They
will still fail with this fix, but with a better error message.


kendase3:
gr8

PR #232, opened 2017-09-19
Chromeover users must now select an image

kendase3:
Addresses OVER-5415

Also reworks validatePage logic for clarity across versions.

### DIFF
--- a/src/image_select_page.cc
+++ b/src/image_select_page.cc
@@ -75,7 +75,7 @@ bool ImageSelectPage::validatePage() {
   }
   // currently this is only a concern in the chromeover case, but we would
   // be equally worried were this true in either case
-  if (bitnessButtons.checkedButton() == NULL) {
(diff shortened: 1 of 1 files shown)


src/image_select_page.cc
```diff
@@ -68,16 +68,21 @@ void ImageSelectPage::initializePage() {
...
-  } else if (hasError) {
+  // if there is an error, we need to allow the user to proceed to the error
+  // screen
+  if (hasError) {
     return true;
-  } else {
+  }
+  // currently this is only a concern in the chromeover case, but we would
+  // be equally worried were this true in either case
+  if (bitnessButtons.checkedButton() == NULL) {
```
nicholasbishop: Prefer `nullptr` to `NULL` macro, or just use `if (!bitnessButtons.checkedButton()) {`

PR #235, opened 2017-09-26
Fixed type

dgs3:
I think this is necessary?


kendase3:
oh yeah, that was a change made in PR after testing.  LG

PR #239, opened 2017-09-28
Kendall 1.9 hotfix 01

kendase3:
My current solution involves chopper plopperin' the constructor I would like to use, as even though the method I want to override is virtual, calls to it from the parent constructor do not call my child method.

### DIFF
--- a/src/gpt_pal.cc
+++ b/src/gpt_pal.cc
@@ -37,8 +37,8 @@ bool clearMbrGpt(const char* physical_path) {
   std::string physical_path_str(physical_path);
   PalData gptdata;
   gptdata.LoadPartitions(std::string(physical_path));
-  // attempt to fix any gpt/mbr problems by setting to a sane, empty state
   int quiet = true;
+  // attempt to fix any gpt/mbr problems by setting to a sane, empty state
   gptdata.SaveGPTData(quiet);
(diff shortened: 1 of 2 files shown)


nicholasbishop:
As discussed IRL this can be simplified by using the other GPTData constructor

kendase3:
comend me

src/gpt_pal.cc
```diff
@@ -19,9 +19,24 @@
...
+  // The disk may be in a weird state, but we are about to reformat it.
+  // Just always use a new partition table.
+  return use_gpt;
+}
+
 bool clearMbrGpt(const char* physical_path) {
   std::string physical_path_str(physical_path);
-  GPTData gptdata(physical_path_str);
+  PalData gptdata;
+  gptdata.LoadPartitions(std::string(physical_path));
```
nicholasbishop: Did it turn out that this call is actually needed despite the fact we are about to overwrite the partitions? If so pls add a comment above

kendase3: i think it's needed to set the disk; it's the only way the GPTData structure knows what to operate on.  It has contents like "device = deviceFilename;"  although now that you mention it; it could be we need fewer subsequent calls.

PR #242, opened 2017-10-04
Add an integration test for general wizard flow

kendase3:
We stub out the download of the image, the unzip step, and the disk image write.  It runs as part of the Linux build.  I touched a bunch of places to allow for the creation of extended stub objects.

Partially addresses OVER-4947

### DIFF
--- a/linuxtest
+++ b/linuxtest
@@ -1,0 +1,2 @@
+#! /bin/bash
+docker build --file docker/gondar-linux.Dockerfile --tag linux_tests .
--- a/linuxtest
+++ b/linuxtest
@@ -1,2 +1,0 @@
-#! /bin/bash
-docker build --file docker/gondar-linux.Dockerfile --tag linux_tests .
--- a/src/device_picker.h
+++ b/src/device_picker.h
@@ -17,11 +17,13 @@
 #define SRC_DEVICE_PICKER_H_
 
 #include <QButtonGroup>
+#include <QRadioButton>
 #include <QVBoxLayout>
 #include <QWidget>
 
 #include "device.h"
 #include "option.h"
+#include "util.h"
 
 namespace gondar {
 
@@ -38,15 +40,32 @@ class DevicePicker : public QWidget {
  signals:
   void selectionChanged();
 
- private:
+ protected:
   class Button;
 
-  const Button* selectedButton() const;
+  virtual const Button* selectedButton() const;
 
   void onButtonClicked(QAbstractButton* button);
 
   QButtonGroup button_group_;
   QVBoxLayout layout_;
+};
+
+class DevicePicker::Button : public QRadioButton {
+ public:
+  Button(const DeviceGuy& device, QWidget* parent)
+      : QRadioButton(QString::fromStdString(device.name), parent),
+        device_(device) {
+    if (device.num_bytes < 6 * gondar::getGigabyte()) {
+      setEnabled(false);
+      setText(QString::fromStdString(device.name) + " (too small)");
+    }
+  }
+
+  const DeviceGuy& device() const { return device_; }
+
+ private:
+  DeviceGuy device_;
 };
 
 }  // namespace gondar
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -29,11 +29,13 @@
 
 class GondarWizard::Private {
  public:
+  Private();
+  Private(gondar::DevicePicker* picker_in);
   gondar::UpdateCheck updateCheck;
   gondar::AboutDialog aboutDialog;
 
   AdminCheckPage adminCheckPage;
-  DeviceSelectPage deviceSelectPage;
+  DeviceSelectPage* deviceSelectPage;
   ChromeoverLoginPage chromeoverLoginPage;
   SiteSelectPage siteSelectPage;
   ErrorPage errorPage;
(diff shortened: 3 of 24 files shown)


kendase3:
^ lil' fixins

test/test.cc
```diff
@@ -62,6 +138,36 @@ void Test::testDevicePicker() {
   QCOMPARE(*picker.selectedDevice(), DeviceGuy(3, "c", getValidDiskSize()));
 }
 
+void proceed(GondarWizard* wizard) {
+  // occassionally with 3 seconds (3k ms) i get stuck on page 3.
```
nicholasbishop: occasionally

linuxtest
```diff
@@ -0,0 +1,2 @@
+#! /bin/bash
+docker build --file docker/gondar-linux.Dockerfile --tag linux_tests .
```
nicholasbishop: I think this is basically the same as running `make jenkins-linux`

test/test.cc
```diff
@@ -35,8 +110,8 @@ QAbstractButton* getDevicePickerButton(DevicePicker* picker, const int index) {
 }  // namespace
 
 uint64_t getValidDiskSize() {
-    const uint64_t gigabyte = 1073741824LL;
-    return 10 * gigabyte;
+  const uint64_t gigabyte = 1073741824LL;
+  return 10 * gigabyte;
```
nicholasbishop: These lil style fixins already got merged to master, maybe a merge from master to this branch will get the diff a lil smaller?

test/test.cc
```diff
@@ -62,6 +138,36 @@ void Test::testDevicePicker() {
   QCOMPARE(*picker.selectedDevice(), DeviceGuy(3, "c", getValidDiskSize()));
 }
 
+void proceed(GondarWizard* wizard) {
+  // occassionally with 3 seconds (3k ms) i get stuck on page 3.
+  QTest::mouseClick(wizard->button(QWizard::NextButton), Qt::LeftButton,
+                    Qt::NoModifier, QPoint(), 3000);
+  LOG_WARNING << "id after click=" << wizard->currentId();
```
nicholasbishop: Is "clicking" the next button meaningfully different from calling `GondarWizard::next()`?

test/test.h
```diff
@@ -17,16 +17,73 @@
...
+  void run() override;
+};
+
+class TestWriteOperationPage : public WriteOperationPage {
+ public:
+  TestWriteOperationPage(QWidget* parent = 0);
+  DiskWriteThread* makeDiskWriteThread(DeviceGuy* drive_in,
+                                       const QString& image_path_in,
+                                       QObject* parent) override;
+};
```
nicholasbishop: These classes probably don't need to be declared in the header, can inline them into the `.cc` file as long as the MOC doesn't need to do anything with them.

src/device_picker.h
```diff
@@ -38,17 +40,34 @@ class DevicePicker : public QWidget {
...
-  const Button* selectedButton() const;
+  virtual const Button* selectedButton() const;
 
   void onButtonClicked(QAbstractButton* button);
 
   QButtonGroup button_group_;
   QVBoxLayout layout_;
 };
 
+class DevicePicker::Button : public QRadioButton {
```
nicholasbishop: Why'd this class move to the header?

test/test.cc
kendase3: i suspect it tests that the button is currently selectable on the page and is a slightly larger-scoped test than just calling next(), but have not tested for differences

src/device_picker.h
kendase3: it's referred to in test.cc

test/test.h
kendase3: i get a moc error when i move them.  supposedly i can add SKIP_AUTOMOC for the file if it's a big deal

nicholasbishop: Maybe you left some unnecessary `Q_OBJECT` macros in then? Those shouldn't be necessary unless you have signals (or require some of the runtime-type info stuff where it provides stringly-typed names of classes/methods)

test/test.cc
```diff
@@ -63,6 +151,36 @@ void Test::testDevicePicker() {
   QCOMPARE(*picker.selectedDevice(), DeviceGuy(3, "c", getValidDiskSize()));
 }
 
+void proceed(GondarWizard* wizard) {
+  // occasionally with 3 seconds (3k ms) i get stuck on page 3.
```
nicholasbishop: The flakiness is weird, any idea what causes it?

kendase3: We wait for newestImageUrl.isReady().  That likely doesn't always return in time.  I could biggify this value.

src/gondarwizard.cc
```diff
@@ -29,11 +29,13 @@
...
 class GondarWizard::Private {
  public:
+  Private();
+  Private(gondar::DevicePicker* picker_in);
   gondar::UpdateCheck updateCheck;
   gondar::AboutDialog aboutDialog;
 
   AdminCheckPage adminCheckPage;
-  DeviceSelectPage deviceSelectPage;
+  DeviceSelectPage* deviceSelectPage;
```
nicholasbishop: I think this should be an owned pointer (`std::unique_ptr`)

test/test.cc
nicholasbishop: Can we stub that out? IMO network IO shouldn't happen during this test

kendase3: this transaction is one of the few substantive things this test does atm.  i'll stub qq

nicholasbishop: Before you do, would it make sense to change the mock to the network level? I.e. rather than stub out NewestImageUrl, create a GondarNetworkManager that can be mocked to return custom JSON for the test / custom installer image for the download part

kendase3: i stubbed out newestimageurl for now as that seemed more consistent with the approach used throughout the test

PR #244, opened 2017-10-05
embiggen font for OVER-5347

kendase3:
This may end up being too big.  Now that we have a means of changing it and we don't like the default, it will be easy to toy around with the constant.

### DIFF
--- a/src/main.cc
+++ b/src/main.cc
@@ -32,6 +32,10 @@ int main(int argc, char* argv[]) {
   gondar::SendMetric(gondar::Metric::Use);
   QApplication app(argc, argv);
   app.setStyleSheet(gondar::readUtf8File(":/style.css"));
+  // embiggen font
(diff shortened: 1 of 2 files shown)


src/main.cc
```diff
@@ -32,6 +32,10 @@ int main(int argc, char* argv[]) {
   gondar::SendMetric(gondar::Metric::Use);
   QApplication app(argc, argv);
   app.setStyleSheet(gondar::readUtf8File(":/style.css"));
+  // embiggen font
+  QFont font;
+  font.setPointSize(font.pointSize() + 4);
+  app.setFont(font);
```
nicholasbishop: Might be better to do this in the style sheet

kendase3: shore

nicholasbishop:
# LGTM

PR #246, opened 2017-10-05
Add a test with minimal stubbing (bigtest)

kendase3:
This test allows us to breeze by the device picker screen, but otherwise exercises everything but the disk write step.
Partially addresses OVER-4947
This new test won't be run on every build.


kendase3:
disregard

PR #247, opened 2017-10-05
Add a test with minimal stubbing

kendase3:
This test allows us to breeze by the device picker screen, but otherwise exercises everything but the disk write step.

This new bigtest won't be run on every build.

Partially addresses OVER-4947


kendase3:
also worth noting that this test won't run on docker images atm

kendase3:
note that this does not include any docker-level changes.  it will be run manually via docker run on the jenkins builders when making a release build

test/slow_test.cc
```diff
@@ -0,0 +1,98 @@
...
+      // the case in which none were enabled
+      return NULL;
+    }
+  }
+};
+
+}  // namespace
+
+void proceed(GondarWizard* wizard, int ms = 500);
+void proceed(GondarWizard* wizard, int ms) {
```
nicholasbishop: Prototype isn't needed, the default value can be used in the definition

test/slow_test.cc
```diff
@@ -0,0 +1,98 @@
...
+  // occasionally with 3 seconds (3k ms) i get stuck on page 3.
+  QTest::mouseClick(wizard->button(QWizard::NextButton), Qt::LeftButton,
+                    Qt::NoModifier, QPoint(), ms);
+  LOG_WARNING << "id after click=" << wizard->currentId();
```
nicholasbishop: `LOG_WARNING` -> `LOG_INFO`

And maybe `wizard->currentId()` -> `wizard->currentPage()->title()` so it's human readable?

test/slow_test.cc
```diff
@@ -0,0 +1,98 @@
...
+}
+
+void Test::testDownloadFlow() {
+  initResource();
+  gondar::InitializeLogging();
+  MockDevicePicker* testpicker = new MockDevicePicker();
+  GondarWizard wizard(testpicker);
+  wizard.show();
+  QTest::qWait(1000);
+  // 0->3
```
nicholasbishop: Suggest replacing these numbers (and all the ones below) with page titles

test/slow_test.cc
```diff
@@ -0,0 +1,98 @@
...
+  }
+};
+
+}  // namespace
+
+void proceed(GondarWizard* wizard, int ms = 500);
+void proceed(GondarWizard* wizard, int ms) {
+  // occasionally with 3 seconds (3k ms) i get stuck on page 3.
+  QTest::mouseClick(wizard->button(QWizard::NextButton), Qt::LeftButton,
+                    Qt::NoModifier, QPoint(), ms);
```
nicholasbishop: From the docs:

> If delay is specified, the test will wait for the specified amount of milliseconds before pressing and before releasing the button.

I think we only really want the delay prior to pressing the button, not a click-and-hold behavior. So maybe set this delay to zero and add a `qWait()` above `mouseClick`

test/slow_test.cc
```diff
@@ -0,0 +1,98 @@
...
+  LOG_WARNING << "id after click=" << wizard->currentId();
+}
+
+void Test::testDownloadFlow() {
+  initResource();
+  gondar::InitializeLogging();
+  MockDevicePicker* testpicker = new MockDevicePicker();
```
nicholasbishop: We enabled c++14 earlier to gain the `optional` class, so the manual memory allocation can be replaced with `make_unique` here. Also no need to duplicate the type name:

```c++
auto picker = std::make_unique<MockDevicePicker>()
```

nicholasbishop:
With the latest change the test crashes for me. I think the device picker ownership needs to move into gondarwizard so that you don't doublefree. Probably just changing parameters to `std::unique_ptr` instead of a raw pointer will fix the issue.

src/device_picker.h
```diff
@@ -38,17 +40,35 @@ class DevicePicker : public QWidget {
...
 
   void onButtonClicked(QAbstractButton* button);
 
   QButtonGroup button_group_;
   QVBoxLayout layout_;
 };
 
+class DevicePicker::Button : public QRadioButton {
+ public:
+  Button(const DeviceGuy& device, QWidget* parent)
```
nicholasbishop: The method definition should move back into the `.cc` file.

src/device_select_page.cc
```diff
@@ -18,35 +18,45 @@
 #include "gondarwizard.h"
 #include "log.h"
 
-DeviceSelectPage::DeviceSelectPage(QWidget* parent) : WizardPage(parent) {
+DeviceSelectPage::DeviceSelectPage(QWidget* parent)
+    : WizardPage(parent), picker(new gondar::DevicePicker()) {
```
nicholasbishop: `std::make_unique`

src/device_select_page.cc
```diff
@@ -18,35 +18,45 @@
...
   // this page should just say 'hi how are you' while it stealthily loads
   // the usb device list.  or it could ask you to insert your device
+  init();
+}
+
+DeviceSelectPage::DeviceSelectPage(gondar::DevicePicker* picker_in,
```
nicholasbishop: `std::unique_ptr`

src/device_select_page.cc
```diff
@@ -18,35 +18,45 @@
...
+                                   QWidget* parent)
+    : WizardPage(parent) {
+  picker.reset(picker_in);
```
nicholasbishop: Rather than have two separate constructors, just have it do the right thing if `picker_in` isn't set. Something like:

```c++
DeviceSelectPage::DeviceSelectPage(std::unique_ptr<gondar::DevicePicker> picker_in, QWidget* parent)
    : WizardPage(parent), picker(picker_in) {
  if (!picker) {
    picker = std::make_unique<gondar::DevicePicker>();
  }
}
```

src/gondarwizard.cc
```diff
@@ -43,10 +45,28 @@ class GondarWizard::Private {
   QDateTime runTime;
 };
 
+GondarWizard::Private::Private() {
+  deviceSelectPage.reset(new DeviceSelectPage());
+}
+GondarWizard::Private::Private(gondar::DevicePicker* picker_in) {
+  deviceSelectPage.reset(new DeviceSelectPage(picker_in));
+}
```
nicholasbishop: I'd recommend not adding methods to the `Private` class. Its members are all public to `GondarWizard`.

src/gondarwizard.cc
```diff
@@ -43,10 +45,28 @@ class GondarWizard::Private {
...
+
 GondarWizard::GondarWizard(QWidget* parent)
     : QWizard(parent),
       p_(new Private()),
       about_shortcut_(QKeySequence::HelpContents, this) {
+  init();
+}
```
nicholasbishop: I think if the suggestions I made above are made then this constructor won't be needed anymore. Incidentally, if you ever do need a constructor that mostly duplicates another constructor, you can delegate to the other constructor with more or less the same syntax as calling a constructor of a parent class.

src/gondarwizard.cc
```diff
@@ -29,11 +29,13 @@
...
 class GondarWizard::Private {
  public:
+  Private();
+  Private(gondar::DevicePicker* picker_in);
   gondar::UpdateCheck updateCheck;
   gondar::AboutDialog aboutDialog;
 
   AdminCheckPage adminCheckPage;
-  DeviceSelectPage deviceSelectPage;
+  std::unique_ptr<DeviceSelectPage> deviceSelectPage;
```
nicholasbishop: I don't think `deviceSelectPage` needs to be a pointer anymore

CMakeLists.txt
```diff
@@ -141,9 +141,13 @@ target_link_libraries(cloudready-usb-maker app)
 
 # Test application
 add_executable(tests test/test.cc)
+add_executable(slowtests test/slow_test.cc)
```
nicholasbishop: To simplify things a little further we could keep the tests in a single `tests` target rather than having two separate executables. The `main()` function that is added automatically by QTest lets you pass in the test(s) that you actually want to run

src/gondarwizard.cc
kendase3: could you go into more detail on this?

CMakeLists.txt
kendase3: so right now i run 'make test' to run fast tests and 'make slowtest' to run the longer ones.  what kind of flow are you suggesting?

kendase3:
i've addressed the unneeded predeclaration of proceed which i thought i had done on monday, but apparently had not

CMakeLists.txt
nicholasbishop: As discussed in meatspace we can defer this

test/slow_test.cc
```diff
@@ -0,0 +1,96 @@
...
+  Q_INIT_RESOURCE(gondarwizard);
+}
+
+namespace gondar {
+
+namespace {
+
+class MockDevicePicker : public gondar::DevicePicker {
+ private:
+  const DevicePicker::Button* selectedButton() const override {
```
nicholasbishop: I'm having some difficulty understanding how this mock picker works. Clearly it does because the test is passing, but I just don't get what it is doing. Add comment please

PR #250, opened 2017-10-19
add What is next? section to final screen

kendase3:
addresses OVER-5495

I will also likely need to either shrink the font size or increase the starting window size.  I haven't tested on winders yet.

### DIFF
--- a/src/write_operation_page.cc
+++ b/src/write_operation_page.cc
@@ -71,15 +71,21 @@ void WriteOperationPage::showWhatsNext() {
   bolded.setText("<br>What's next?<br>");
   layout.addWidget(&bolded);
 
-  //TODO: update links
+  // TODO: update links
   whatsNext.setObjectName("whatsNext");
-  whatsNext.setText("<p>You're ready to install CloudReady!  Head back to <a href=\"https://guide.neverware.com/supported-devices\">the install guide</a> for help in how to use your USB installer.<br></p><p>Don't forget to check the 'Details' link for your devices on the <a href=\"https://guide.neverware.com/supported-devices\">certified models list</a>.  There may be special install instructions or important notes for each model.</p>");
+  whatsNext.setText(
+      "<p>You're ready to install CloudReady!  Head back to <a "
+      "href=\"https://guide.neverware.com/supported-devices\">the install "
+      "guide</a> for help in how to use your USB installer.<br></p><p>Don't "
+      "forget to check the 'Details' link for your devices on the <a "
+      "href=\"https://guide.neverware.com/supported-devices\">certified models "
+      "list</a>.  There may be special install instructions or important notes "
+      "for each model.</p>");
   whatsNext.setTextFormat(Qt::RichText);
   whatsNext.setTextInteractionFlags(Qt::TextBrowserInteraction);
   whatsNext.setOpenExternalLinks(true);
   whatsNext.setWordWrap(true);
   layout.addWidget(&whatsNext);
-
 }
 
 void WriteOperationPage::onDoneWriting() {
--- a/src/write_operation_page.cc
+++ b/src/write_operation_page.cc
@@ -75,7 +75,7 @@ void WriteOperationPage::showWhatsNext() {
   whatsNext.setObjectName("whatsNext");
   whatsNext.setText(
       "<p>You're ready to install CloudReady!  Head back to <a "
-      "href=\"https://guide.neverware.com/supported-devices\">the install "
+      "href=\"https://guide.neverware.com/install-cloudready\">the install "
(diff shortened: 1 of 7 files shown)


src/write_operation_page.cc
```diff
@@ -64,6 +64,31 @@ void WriteOperationPage::showProgress() {
...
 
+void WriteOperationPage::showWhatsNext() {
+  setTitle("CloudReady USB created!");
+  setSubTitle("You may now either exit or create another USB.");
+  whatsNext1.setText("<br>What's next?<br>");
+  whatsNext1.setStyleSheet("font-weight: bold");
+  layout.addWidget(&whatsNext1);
+
+  //TODO: update links
+  whatsNext2.setText("<p>You're ready to install CloudReady!  Head back to <a href=\"https://guide.neverware.com/supported-devices\">the install guide</a> for help in how to use your USB installer.<br></p>");
```
nicholasbishop: Clang format

src/write_operation_page.cc
```diff
@@ -64,6 +64,31 @@ void WriteOperationPage::showProgress() {
...
+  whatsNext2.setTextFormat(Qt::RichText);
+  whatsNext2.setTextInteractionFlags(Qt::TextBrowserInteraction);
+  whatsNext2.setOpenExternalLinks(true);
+  whatsNext2.setWordWrap(true);
+  whatsNext2.setStyleSheet("padding-left:3em");
+  layout.addWidget(&whatsNext2);
+
+  whatsNext3.setText("<p>Don't forget to check the 'Details' link for your devices on the <a href=\"https://guide.neverware.com/supported-devices\">certified models list</a>.  There may be special install instructions or important notes for each model.</p>");
```
nicholasbishop: Since this is rich text there's no need to have separate widgets; just put all three whatsNext into a single label.

src/write_operation_page.cc
```diff
@@ -64,6 +64,31 @@ void WriteOperationPage::showProgress() {
   progress.setValue(0);
 }
 
+void WriteOperationPage::showWhatsNext() {
+  setTitle("CloudReady USB created!");
+  setSubTitle("You may now either exit or create another USB.");
+  whatsNext1.setText("<br>What's next?<br>");
+  whatsNext1.setStyleSheet("font-weight: bold");
```
nicholasbishop: The stylesheet stuff should move to `style.css`

src/write_operation_page.cc
```diff
@@ -64,6 +64,24 @@ void WriteOperationPage::showProgress() {
...
 
+void WriteOperationPage::showWhatsNext() {
+  setTitle("CloudReady USB created!");
+  setSubTitle("You may now either exit or create another USB.");
+  bolded.setText("<br>What's next?<br>");
+  bolded.setStyleSheet("font-weight: bold");
+  layout.addWidget(&bolded);
+
+  //TODO: update links
+  whatsNext.setText("<p>You're ready to install CloudReady!  Head back to <a href=\"https://guide.neverware.com/supported-devices\">the install guide</a> for help in how to use your USB installer.<br></p><p>Don't forget to check the 'Details' link for your devices on the <a href=\"https://guide.neverware.com/supported-devices\">certified models list</a>.  There may be special install instructions or important notes for each model.</p>");
```
nicholasbishop: Can this label be combined with the `bolded` label?

nicholasbishop: clang-format

kendase3: Wouldn't that work against your previous request to move css to the external file?  How would one have bold text and the other be indented without inline css if they're the same entity?

nicholasbishop: I was assuming that Qt style sheets would allow CSS selectors to span both widgets and the HTML subset that labels support, something like this:

```css
#whatsNext p {
    padding-left: 10px;
}
```

But it seems that isn't allowed.

src/write_operation_page.cc
```diff
@@ -64,6 +64,30 @@ void WriteOperationPage::showProgress() {
...
+  setSubTitle("You may now either exit or create another USB.");
+  bolded.setObjectName("bolded");
+  bolded.setText("<br>What's next?<br>");
+  layout.addWidget(&bolded);
+
+  // TODO: update links
+  whatsNext.setObjectName("whatsNext");
+  whatsNext.setText(
+      "<p>You're ready to install CloudReady!  Head back to <a "
+      "href=\"https://guide.neverware.com/supported-devices\">the install "
```
nicholasbishop: https://guide.neverware.com/install-cloudready/

kendase3: yeah i just wanted to check with forrest first to see if we wanted any tracking on the real links.  imma update them

PR #255, opened 2017-12-15
Fix "what's next" prompt for 'make another usb' case

kendase3:
Addresses OVER-5567

I have yet to test this on a winders machine.


kendase3:
I tested this on a winders machine.

PR #257, opened 2017-12-28
Include site id in metrics when available

kendase3:
Addresses OVER-5622

Feel free to suggest an alternative to my static member method for storing state in the metrics layer.

### DIFF
--- a/src/metric.cc
+++ b/src/metric.cc
@@ -116,7 +116,7 @@ void SetSiteId(int site_id) {
   HoldSiteId(site_id);
 }
 
-int GetSiteId() {
+static int GetSiteId() {
   return HoldSiteId(0);
 }
 
@@ -151,7 +151,7 @@ void SendMetric(Metric metric, const std::string& value) {
   json.insert("product", product);
   const auto siteId = GetSiteId();
   // only show site when on beerover and site id has been initialized
-  if ((isChromeover() && siteId != 0) {
+  if (isChromeover() && siteId != 0) {
     json.insert("site", siteId);
   }
   QNetworkRequest request(url);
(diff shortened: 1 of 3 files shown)


kendase3:
self-assigning for a moment; one of my refactors broke the syntax here

src/metric.cc
```diff
@@ -131,6 +149,11 @@ void SendMetric(Metric metric, const std::string& value) {
     product = "beerover";
   }
   json.insert("product", product);
+  const auto siteId = GetSiteId();
+  // only show site when on beerover and site id has been initialized
```
tedbrandston: > only show site when on beerover

src/metric.cc
```diff
@@ -102,6 +102,24 @@ static QString getUuid() {
...
+  return site_id;
+}
+
+void SetSiteId(int site_id) {
+  HoldSiteId(site_id);
+}
+
+static int GetSiteId() {
+  return HoldSiteId(0);
+}
```
tedbrandston: So, yeah, this whole thing strikes me as a little weird, but not totally out of the question.

I'm assuming you considered keeping the site info in the same way you track other data page-to-page and then adding it as an (optional?) param to SendMetric?

I'm also guessing you're wrapping the actual static int in a function to make it harder for other stuff to mess with it? I'm pretty sure you can instead put it into an anonymous namespace and get the same effect (though it's been a little while since I did something like that). I think that would be clearer:
```c++
namespace {
static int CURRENT_SITE_ID = 0;
}

void SetSiteId(int site_id) {
  CURRENT_SITE_ID = site_id;
}

static int GetSiteId() {
  return CURRENT_SITE_ID;
}
```

I think that's usually how I see things like that done?

kendase3:
@tedbrandston take another look when you get a chance

src/metric.cc
kendase3: i believe the above is more accurate; note that some metrics will fire before a user has entered their credentials and then selected a site.

PR #260, opened 2018-09-01
Now inform user of the problem if they have zero sites

kendase3:
Addresses OVER-7203


src/chromeover_login_page.cc
```diff
@@ -93,6 +93,12 @@ void ChromeoverLoginPage::handleMeepoFinished() {
     // TODO: let's make this a commit page so we don't have to worry about
     // the case of a user revisiting this segment for now
     wizard()->next();
+  // If the user has no sites, proceed to the error screen with the
+  // appropriate error
+  } else if (QString::compare(meepo_.no_sites_error, meepo_.error()) == 0) {
```
nicholasbishop: Can probably just directly compare these strings, no need for explicit `QString::compare`

PR #262, opened 2018-09-20
update this file to latest version used in rufus

kendase3:
This is required for OVER-6235 (suppression of windows popups)


kendase3:
alternatively i could pick and choose functionality from this file, but that could get hairier if/when we want additional functionality from rufus later.

PR #263, opened 2018-09-20
Use code from current version of rufus to dismiss windows format popups

kendase3:
TODO: work for different (non-english) locales

Addresses OVER-6235

I still have some cleaning up to do but YGtI

### DIFF
--- a/src/dismissprompt.h
+++ b/src/dismissprompt.h
@@ -3,13 +3,6 @@
 #define DISMISSPROMPT_H_
 
 #include <windows.h>
-#include <windowsx.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <string.h>
-#include <shlobj.h>
-#include <commdlg.h>
-#include <richedit.h>
 
 #include "msapi_utf8.h"
 
--- a/src/dismissprompt.h
+++ b/src/dismissprompt.h
@@ -2,14 +2,14 @@
 #ifndef DISMISSPROMPT_H_
 #define DISMISSPROMPT_H_
 
-#include <windows.h>
-#include <windowsx.h>
-#include <stdlib.h>
-#include <stdio.h>
-#include <string.h>
-#include <shlobj.h>
 #include <commdlg.h>
 #include <richedit.h>
+#include <shlobj.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <windows.h>
+#include <windowsx.h>
 
 #include "msapi_utf8.h"
 
--- a/src/dismissprompt.h
+++ b/src/dismissprompt.h
@@ -2,14 +2,14 @@
 #ifndef DISMISSPROMPT_H_
 #define DISMISSPROMPT_H_
 
-#include <commdlg.h>
-#include <richedit.h>
-#include <shlobj.h>
-#include <stdio.h>
-#include <stdlib.h>
-#include <string.h>
 #include <windows.h>
 #include <windowsx.h>
+#include <stdlib.h>
+#include <stdio.h>
+#include <string.h>
+#include <shlobj.h>
+#include <commdlg.h>
+#include <richedit.h>
 
 #include "msapi_utf8.h"
 
--- a/src/dismissprompt.h
+++ b/src/dismissprompt.h
@@ -1,0 +1,18 @@
+
+#ifndef DISMISSPROMPT_H_
+#define DISMISSPROMPT_H_
+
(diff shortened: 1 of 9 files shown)


src/shared.h
```diff
@@ -60,4 +60,8 @@ files
 
 #define safe_strdup _strdup
 
+// from bled
```
nicholasbishop: What's bled?

kendase3: it's a subproject of rufus

src/dismissprompt.h
```diff
@@ -0,0 +1,18 @@
...
+#define DISMISSPROMPT_H_
+
+#include <commdlg.h>
+#include <richedit.h>
+#include <shlobj.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <windows.h>
+#include <windowsx.h>
```
nicholasbishop: This seems like a lot of includes for an almost-empty header file, can some of these go to the .cc file?

src/msapi_utf8.h
```diff
@@ -116,16 +226,17 @@ static __inline HANDLE CreateFileU(const char* lpFileName,
...
+static __inline BOOL SetupDiGetDeviceRegistryPropertyU(
     HDEVINFO DeviceInfoSet,
     PSP_DEVINFO_DATA DeviceInfoData,
     DWORD Property,
     PDWORD PropertyRegDataType,
     PBYTE PropertyBuffer,
     DWORD PropertyBufferSize,
     PDWORD RequiredSize) {
-  bool ret = false;
+  BOOL ret = FALSE;
```
nicholasbishop: These bool changes seem unnecessary

src/msapi_utf8.h
```diff
@@ -23,16 +23,49 @@
...
+  }
+
+// Never ever use isdigit() or isspace(), etc. on UTF-8 strings!
+// These calls take an int and char is signed so MS compilers will produce an
+// assert error on anything that's > 0x80
+#define isasciiU(c) isascii((unsigned char)(c))
+#define iscntrlU(c) iscntrl((unsigned char)(c))
+#define isdigitU(c) isdigit((unsigned char)(c))
+#define isspaceU(c) isspace((unsigned char)(c))
+#define isxdigitU(c) isxdigit((unsigned char)(c))
```
nicholasbishop: Are any of these used?

kendase3:
closing for now; i frogot to separate out this code into the windows-specific implementation.  i'll have to think about the cleanest way to do that (i had it as part of the lib rather than the app).

kendase3:
i'm testing this out now but it should address the windows/linux thing

kendase3:
so i've tested this out and the only corner case i've found is that if you already had a windows format popup from before you launched gondar and you never focus on that window until you close gondar, it will still be there.  that seems like an ok corner case for now.  i'm going to work on the translation bit next.

kendase3:
retest this please

kendase3:
i've made separate tickets for the non-english support and the corner case mentioned above

src/msapi_utf8.h
```diff
@@ -33,6 +34,24 @@ extern "C" {
...
+#define GetSaveFileNameU(p) GetOpenSaveFileNameU(p, TRUE)
+#define GetOpenFileNameU(p) GetOpenSaveFileNameU(p, FALSE)
+#define ListView_SetItemTextU(hwndLV, i, iSubItem_, pszText_)           \
+  {                                                                     \
+    LVITEMW _ms_wlvi;                                                   \
+    _ms_wlvi.iSubItem = iSubItem_;                                      \
+    _ms_wlvi.pszText = utf8_to_wchar(pszText_);                         \
+    SNDMSG((hwndLV), LVM_SETITEMTEXTW, (WPARAM)(i), (LPARAM)&_ms_wlvi); \
+    sfree(_ms_wlvi.pszText);                                            \
+  }
```
nicholasbishop: Are these additions used?

src/msapi_utf8.h
```diff
@@ -96,6 +115,81 @@ static __inline wchar_t* utf8_to_wchar(const char* str) {
   return wstr;
 }
 
+/*
+* Converts an non NUL-terminated UTF-16 string of length len to UTF8 (allocate
+* returned string)
+* Returns NULL on error
+*/
+static __inline char* wchar_len_to_utf8(const wchar_t* wstr, int wlen) {
```
nicholasbishop: Is this function used?

PR #265, opened 2018-10-01
Kendall warp tunnel 13

kendase3:
I'm still cleaning up some warnings but it works in this state.  Presented for the approval of the midnight society...


src/admin_check_page.h
```diff
@@ -36,6 +37,8 @@ class AdminCheckPage : public gondar::WizardPage {
 
  private:
   QLabel label;
+  // offer the users a way to format their disk
+  QLabel warpTunnel;
```
nicholasbishop: How about `format_disk` or something like that?

src/mkfs_pal.h
```diff
@@ -0,0 +1,7 @@
+
+#ifndef MKFS_PAL_H
```
nicholasbishop: How about just `mkfs` instead of `mkfs_pal`?

src/mkfs_pal.cc
```diff
@@ -0,0 +1,120 @@
...
+    LOG_WARNING << "unknown callback case";
+    break;
+  }
+  return true;
+}
+
+
+void makeFilesystem(char* logical_path) {
+  LOG_WARNING << "making filesystem...";
+  // LoadLibrary("fmifs.dll") appears to changes the locale, which can lead to
```
nicholasbishop: changes -> change

src/write_operation_page.cc
```diff
@@ -87,10 +94,40 @@ void WriteOperationPage::showProgress() {
...
+  whatsNext.setTextInteractionFlags(Qt::TextBrowserInteraction);
+  whatsNext.setOpenExternalLinks(true);
+  whatsNext.setWordWrap(true);
+  layout.addWidget(&whatsNext);
+
   bolded.show();
   whatsNext.show();
 }
 
+// fixme: where is the format-only guy supposed to go?
```
nicholasbishop: Not sure what this fixme means

src/write_operation_page.cc
```diff
@@ -68,14 +68,21 @@ bool WriteOperationPage::validatePage() {
 
 void WriteOperationPage::writeToDrive() {
   LOG_INFO << "Writing to drive...";
-  image_path.clear();
-  image_path.append(wizard()->downloadProgressPage.getImageFileName());
-  showProgress();
-  diskWriteThread = new DiskWriteThread(&device, image_path, this);
+  // if we're in warp mode, we don't need any logic about an image file name
```
nicholasbishop: Would it make sense to create a new `Page` for reformatting instead of making it part of `WriteOperationPage`? I'm not sure one way or the other, but it does seem like a lot of the logic is separated in these `if` blocks.

src/msapi_utf8.h
```diff
@@ -25,6 +25,15 @@
...
 #include <setupapi.h>
 #include <windows.h>
+// TODO: necessary?
+#include <setupapi.h>
+#include <usbioctl.h>
+#include <versionhelpers.h>
+
+#include <inttypes.h>
+#include <stdio.h>
+#include <stdlib.h>
```
nicholasbishop: Probly not necessary in the header file, I imagine these can move to `mkfs.cc` if they are needed there

src/mkfs_pal.cc
```diff
@@ -0,0 +1,120 @@
...
+  // problems with tolower(). Make sure we restore the locale. For more details,
+  // see http://comments.gmane.org/gmane.comp.gnu.mingw.user/39300
+  char* locale = setlocale(LC_ALL, NULL);
+  pfFormatEx = (FormatEx_t) GetProcAddress(LoadLibraryA("fmifs.dll"), "FormatEx");
+  setlocale(LC_ALL, locale);
+
+  // TODO: make sure the name i'm passing is this
+  wchar_t* logical_path_windows = utf8_to_wchar(logical_path);
+  pfFormatEx(logical_path_windows, RemovableMedia, L"FAT32", L"", /*quick*/true, /*clustersize*/4096, FormatExCallback);
```
nicholasbishop: A comment about why the cluster size is 4096 would be good.

kendase3: sure; general reasoning from answer found https://superuser.com/questions/1286135/what-allocation-unit-size-to-use-when-formatting-a-usb-flash-drive-in-fat32/1287233

src/write_operation_page.cc
kendase3: that makes sense.  the gui bits of this PR are from like a year ago; i was working out the backend part when i got retasked.  i'll take another look at the gui flow and look at making a separate page.  the 'make another usb' flow does not really make sense with formatting, so that's another reason to separate.

kendase3: hmm actually that case is taken care of; i don't display the make another prompt for warp tunnel case.

kendase3:
I have to make some changes for this to run on Linux.  Otherwise, I believe I've addressed the above feedback.

kendase3:
let me know what you think about this; i could also just add more ifndef branching to main.cc instead of adding to stubs.

src/write_operation_page.cc
```diff
@@ -68,14 +68,21 @@ bool WriteOperationPage::validatePage() {
...
   LOG_INFO << "Writing to drive...";
-  image_path.clear();
-  image_path.append(wizard()->downloadProgressPage.getImageFileName());
-  showProgress();
-  diskWriteThread = new DiskWriteThread(&device, image_path, this);
+  // if we're in format-only mode, we don't need logic about an image file name
+  if (wizard()->isFormatOnly()) {
+    // make a disk write thread in format mode
+    diskWriteThread = new DiskWriteThread(&device, this);
+    // TODO: send a metric about a clear drive attempt
```
kendase3: I can address this TODO in another PR and hopefully squeak into 2.2

kendase3:
@nicholasbishop have another taste when you get a sec

src/diskwritethread.cc
```diff
@@ -30,6 +30,11 @@ static int64_t getFileSize(QString& path) {
   return file.size();
 }
 
+DiskWriteThread::DiskWriteThread(DeviceGuy* drive_in, QObject* parent)
+    : QThread(parent), selected_drive(*drive_in) {
+  image_path = nullptr;
```
nicholasbishop: Setting a QString to null is a little weird; it happens to work because there's an overload of `operator=` for `char*`-style strings, but you can just remove this line and it will be initialized by the default constructor.

src/mkfs.h
```diff
@@ -0,0 +1,8 @@
+
```
nicholasbishop: copyright header

src/diskwritethread.cc
```diff
@@ -65,6 +69,27 @@ void DiskWriteThread::run() {
   setState(State::Success);
 }
 
+void DiskWriteThread::formatDrive() {
+  LOG_INFO << "formatting disk";
+  setState(State::Running);
+  // non-zero is failure
+  if (Format(&selected_drive)) {
+    LOG_ERROR << "Install failed";
+    setState(State::InstallFailed);
```
nicholasbishop: Is this state correct? Seems like it should be FormatFailed

src/diskwritethread.cc
```diff
@@ -65,6 +69,27 @@ void DiskWriteThread::run() {
...
+    return;
+  }
+
+  LOG_INFO << "Format succeeded";
+  setState(State::Success);
+}
+void DiskWriteThread::run() {
+  if (image_path == nullptr) {
```
nicholasbishop: `if (image_path.empty())`

src/gondar.cc
```diff
@@ -2278,21 +2279,29 @@ DeviceGuyList GetDeviceList() {
...
+  return true;
+}
+
 bool Install(DeviceGuy* target_device,
              const char* image_path,
              int64_t image_size) {
   uint64_t device_num = target_device->device_num;
   uint64_t sector_size = GetSectorSize(device_num);
   uint64_t drive_size = GetDriveSize(device_num);
-  // FIXME: this is a leak
```
nicholasbishop: Is this fixed? I didn't see where

src/mkfs.cc
```diff
@@ -0,0 +1,152 @@
...
+
+#include <inttypes.h>
+#include <windows.h>
+#include <winioctl.h>  // for MEDIA_TYPE
+
+#include "log.h"
+#include "msapi_utf8.h"
+
+HMODULE OpenedLibraryHandle = NULL;
+static __inline HMODULE GetLibraryHandle() {
```
nicholasbishop: Unused?

src/mkfs.cc
```diff
@@ -0,0 +1,152 @@
...
+  // see http://comments.gmane.org/gmane.comp.gnu.mingw.user/39300
+  char* locale = setlocale(LC_ALL, NULL);
+  pfFormatEx =
+      (FormatEx_t)GetProcAddress(LoadLibraryA("fmifs.dll"), "FormatEx");
+  setlocale(LC_ALL, locale);
+
+  wchar_t* logical_path_windows = utf8_to_wchar(logical_path);
+  std::wstring fat32str = L"FAT32";
+  std::wstring emptystr = L"";
+  // TODO: is it ok to cast from const wchar * to wchar *?
```
nicholasbishop: Technically no, although as long as the function being called lower down doesn't modify the inputs it should be OK.

src/gondar.cc
kendase3: this is one of the changes i made in the initial version of this PR from like a year ago.  i vaguely recall removing this intentionally; but i'll add it back in for now.

src/diskwritethread.cc
kendase3: There's no FormatFailed state; InstallFailed currently covers all USB writes

src/gondar.cc
```diff
@@ -2327,3 +2337,26 @@ bool Install(DeviceGuy* target_device,
 
   return ret;
 }
+
+bool Format(DeviceGuy* target_device) {
```
nicholasbishop: It's a little confusing that some of these new functions return false on success and true on failure, can we invert that?

src/gpt_pal.cc
```diff
@@ -33,11 +38,33 @@ WhichToUse PalData::UseWhichPartitions(void) {
...
+  partitions[0].SetLastLBA(endSector);
+  partitions[0].SetType(0x0b00);  // make it fat32
+  partitions[0].RandomizeUniqueGUID();
+  DisplayGPTData();
+  // arg is 'quiet'
+  SaveGPTData(true);
+}
+
+// regardless this is the 'shared' logic
+// this looks good now.
```
nicholasbishop: Regardless of what? Not sure what this comment is referring to

src/gondar.cc
kendase3: sure, that's an oversight.  i'll do non-zero on error

src/mkfs.cc
```diff
@@ -0,0 +1,149 @@
...
+#include "mkfs.h"
+
+#include <inttypes.h>
+#include <windows.h>
+#include <winioctl.h>  // for MEDIA_TYPE
+
+#include "log.h"
+#include "msapi_utf8.h"
+
+HMODULE OpenedLibraryHandle = NULL;
```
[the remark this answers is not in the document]
kendase3: good catch; that was dangling from an earlier implementation.  i made that before i realized we were opening more than one library (see gondar.cc)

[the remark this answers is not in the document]
kendase3: i'll put up a fix

kendase3:
the problems i have with things as is:
1) we don't clean up libraries at program termination as rufus does
2) i have two separate implementations of library usage

i'm looking at a way to address this without the implementation looking too bananas

kendase3:
I've fixed my earlier attempt to clean up this library.  I think this way of allocating libraries is quite a bit clearer than the way rufus does it (as i cargo-culted in rufus.cc).  Eventually I'd like to have that one look like this one rather than vice-versa.  The variable names (such as pfFormat) are not particularly meaningful to me and could be changed if you have feelings.

kendase3:
Well, fixed is putting it strongly.  I have yet to test it on a windows machine.

kendase3:
@nicholasbishop take another look at this goldcode if you get a chance

PR #269, opened 2018-10-08
not sure why the 32-bit logic was outside the conditional before

kendase3:


kendase3:
retest this please

kendase3:
apparently jenkins slave 1 is 'being weird'

PR #270, opened 2018-10-09
Add confirmation popup for 32-bit image selection

kendase3:
Addresses OVER-7517

### DIFF
--- a/src/image_select_page.cc
+++ b/src/image_select_page.cc
@@ -83,7 +83,7 @@ bool ImageSelectPage::validatePage() {
     confirmBox.setText(
         "32-bit CloudReady is not supported on 64-bit machines. Use 32-bit "
         "CloudReady only on hardware that requires it.");
-    // counter-intuitive that RejectRole be to go forward, but the roles
+    // counter-intuitive that RejectRole maps to forward, but the roles
     // really just determine button order and are not used for later logic
     QPushButton* backButton =
         confirmBox.addButton("Back", QMessageBox::ActionRole);
--- a/src/image_select_page.cc
+++ b/src/image_select_page.cc
@@ -80,7 +80,7 @@ bool ImageSelectPage::validatePage() {
     confirmBox.setIcon(QMessageBox::Question);
     confirmBox.setWindowTitle("CloudReady");
     confirmBox.setText("32-bit CloudReady is not supported on 64-bit machines. Use 32-bit CloudReady only on hardware that requires it.");
-    QPushButton *backButton = confirmBox.addButton(QMessageBox::Abort);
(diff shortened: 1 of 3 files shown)


src/image_select_page.cc
```diff
@@ -74,6 +74,20 @@ bool ImageSelectPage::validatePage() {
   if (hasError) {
     return true;
   }
+  if (bitnessButtons.checkedButton() ==
+      qobject_cast<QAbstractButton*>(&thirtyTwo)) {
+    switch (QMessageBox::question(
+        this, "CloudReady",
+        "32-bit CloudReady is not supported on 64-bit machines. Use 32-bit "
+        "CloudReady only on hardware that requires it.  Continue?",
+        QMessageBox::No | QMessageBox::Yes)) {
```
nicholasbishop: I prefer buttons to have "action" labels rather than yes/no. Maybe something like "Back" instead of "No" and "Use 32-bit" instead of "Yes"?

kendase3:

src/image_select_page.cc
```diff
@@ -74,6 +74,29 @@ bool ImageSelectPage::validatePage() {
...
   }
+  if (bitnessButtons.checkedButton() ==
+      qobject_cast<QAbstractButton*>(&thirtyTwo)) {
+    QMessageBox confirmBox;
+    confirmBox.setIcon(QMessageBox::Question);
+    confirmBox.setWindowTitle("CloudReady");
+    confirmBox.setText(
+        "32-bit CloudReady is not supported on 64-bit machines. Use 32-bit "
+        "CloudReady only on hardware that requires it.");
+    // counter-intuitive that RejectRole be to go forward, but the roles
```
nicholasbishop: "be to go"

src/image_select_page.cc
```diff
@@ -74,6 +74,31 @@ bool ImageSelectPage::validatePage() {
   if (hasError) {
     return true;
   }
+  if (bitnessButtons.checkedButton() ==
+      qobject_cast<QAbstractButton*>(&thirtyTwo)) {
+    QMessageBox confirmBox;
+    confirmBox.setIcon(QMessageBox::Question);
+    confirmBox.setWindowTitle("CloudReady");
```
nicholasbishop: should either use a more descriptive title like "Confirm 32-bit selection" or just use the same title as the parent application, "CloudReady USB Maker".

src/image_select_page.h
```diff
@@ -18,6 +18,7 @@
 
 #include <QButtonGroup>
 #include <QLabel>
+#include <QMessageBox>
```
nicholasbishop: include can move to source file

PR #272, opened 2018-10-10
Initial changes required for minimal AWS codebuild run

kendase3:


kendase3:
disregard this for now; this was a good proof of concept but i'll want to do it slightly differently.

PR #275, opened 2018-10-10
Some small changes to make codebuild checks work

kendase3:
Addresses OVER-7531

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -1,0 +1,10 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - echo Build started on `date`
+      - make build-gondar test
+  post_build:
+    commands:
+      - echo Build completed on `date`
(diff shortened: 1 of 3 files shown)


buildspec.yml
```diff
@@ -0,0 +1,10 @@
+version: 0.2
+
+phases:
+  build:
+    commands:
+      - echo Build started on `date`
+      - make build-gondar test
```
nicholasbishop: Our Jenkins builder also builds the Windows version, and I think that's the part that Jenkins is actually choking on.

kendase3: this doesn't do that yet.  coming soon.

kendase3: this pr doesn't turn off the jenkins builders; they're already off.

kendase3:
i'm now thinking i'm going to make a codebuild subdir to house these .yml files.  i think we'll have 4.

kendase3:
deprecated by https://github.com/neverware/gondar/pull/278

PR #278, opened 2018-10-12
Changes to allow codebuild to create and run dockerfiles

kendase3:
Addresses OVER-7527 and OVER-7531

### DIFF
--- a/codebuild/buildlinux.yml
+++ b/codebuild/buildlinux.yml
@@ -1,0 +1,17 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - echo Logging in
+      - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
+  build:
+    commands:
+      - echo Build started on `date`
+      - git submodule update --init
+      - docker build -t gondar-linux --file docker/gondar-linux.Dockerfile .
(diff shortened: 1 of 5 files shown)


kendase3:
They just swapped the UI and apparently I can't currently save changes to my gondar-windows target to use my in-aws-built dockerfile.  But these changes worked for the linux one so should be good.  I'm pretty sure with this setup we could have the dockerfiles re-run on every PR if that's what we want.

codebuild/buildlinux.yml
```diff
@@ -0,0 +1,17 @@
+version: 0.2
+
+phases:
+  pre_build:
+    commands:
+      - echo Logging in
+      - $(aws ecr get-login --no-include-email --region $AWS_DEFAULT_REGION)
```
nicholasbishop: Why is this command wrapped in `$(...)`?

kendase3: the aws example did that; maybe to resolve the env vars?

nicholasbishop: Seems strange but I'll live.

kendase3: i'll try without and see what happens

kendase3: nevermind u love it like this

PR #298, opened 2019-01-15
Minimal fix to get MXE dockerfile running again

kendase3:
Our old target no longer builds; we were just using a cached dockerfile.  Updating to latest and adding some packages got MXE building again


kendase3:
note: i don't believe this dockerfile is run for the test above; the cached version is used.

PR #302, opened 2019-02-06
WIP: remove MXE dependency for cross-compilation

kendase3:
I'm not sure if the resultant builds are statically linked yet; I'll have access to a windows machine tomorrow.  Also note that DCMAKE_C_IMPLICIT_INCLUDE_DIRECTORIES should only be set for the windows builds; right now it's unconditional.


test/slow_test.cc
```diff
@@ -21,10 +21,6 @@
 #include "src/gondarwizard.h"
 #include "src/log.h"
 
-#if defined(Q_OS_WIN)
-Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin);
```
nicholasbishop: Why remove this?

kendase3: it seemed as though they reworked their platform lib; my suspicion is it is no longer necessary
i'll find out tomorrow

kendase3: it's also possible that the fedora version of mingw qt is missing this plugin

kendase3: i ran into stuff basically like https://github.com/mxe/mxe/issues/1642#issuecomment-404491987

kendase3: here's where it's suggested linking qwindows should suffice for static windows integration goodies https://bugzilla.redhat.com/show_bug.cgi?id=1529178

kendase3:
i'll put up an updated, more functional version of this idea.  this version is missing some goodies.

PR #303, opened 2019-02-13
Add Google Sign-in to Gondar

kendase3:
This implementation uses the existing MXE container and Fedora 25 environment.  A local webserver is started using libmicrohttpd at a random port between 4000 and 5000.  Users are able to authenticate with Google using their existing browser session.  See TODO/FIXMEs for limitations of current approach.

Addresses OVER-7586


kendase3:
my first suspicion with the build failure above is because i ran yapf on package.py which had an issue in the past.  i've tested this code recently and it works on my machine.  i'll debug the above problem soon.

.gitignore
```diff
@@ -2,5 +2,6 @@
 /package/
 /mxe/
 *.lz4
+*.pyc
```
nicholasbishop: This suggests that we're using python2 somewhere instead of python3?

docker/gondar-win32.Dockerfile
```diff
@@ -2,6 +2,16 @@ FROM neverware/gondar-build-mxe:v3
...
 
+RUN dnf install -y texinfo
+
+ADD libmicrohttpd /opt/gondar/libmicrohttpd
+# build microhttpd first so it's faster to iterate on our own changes
+WORKDIR /opt/gondar/libmicrohttpd
+run ./bootstrap
+run ./configure --host=i686-w64-mingw32.static --enable-static --disable-shared --disable-https
+run make
+run make install
```
nicholasbishop: nitpick: capitalize `run` to match existing style

docker/gondar-win32.Dockerfile
```diff
@@ -2,6 +2,16 @@ FROM neverware/gondar-build-mxe:v3
 
 ENV TREAT_WARNINGS_AS_ERRORS=true
 
+RUN dnf install -y texinfo
```
nicholasbishop: I imagine this is just needed for building docs? Can probably add `--disable-doc` or similar to the configure line and get rid of this.

src/chromeover_login_page.h
```diff
@@ -47,8 +51,21 @@ class ChromeoverLoginPage : public gondar::WizardPage {
...
+  QString client_id;
+  QString state;
+  QString client_secret;
+  QString redirect_uri;
+  QString code_verifier;
+ private slots:
+  void handleGoogleSigninError(QString error);
+  void handleGoogleSigninPart1();
+  void handleGoogleSigninPart2(QString state_in, QString code);
+  void handleGoogleSigninPart3(QNetworkReply* reply);
```
nicholasbishop: Nitpick: `private slots:` isn't actually different from `private:`, so I'd remove that and move these methods up to where `void handleMeepoFinished();` is

src/util.cc
```diff
@@ -24,6 +24,49 @@
...
+  return QByteArray(GOOGLE_SIGN_IN_SECRET);
+#else
+  return QByteArray();
+#endif
+}
+
+// FIXME: currently using running system time as seed for random generator
+// seed our random number generator if it has not yet been seeded
+void maybeInitRand() {
+  static bool run = false;
```
nicholasbishop: Why not just call this once at startup rather than before each time `rand()` is used?

src/util.cc
```diff
@@ -24,6 +24,49 @@
...
+
+// get the port the local google sign in server will run on
+int getPort() {
+  static int port = 0;
+  if (port == 0) {
+    // generate a new port
+    maybeInitRand();
+    int MAX_PORT = 5000;
+    int MIN_PORT = 4000;
+    port = MIN_PORT + std::rand() % (MAX_PORT - MIN_PORT);
```
nicholasbishop: nitpick: in C++11 you can use `uniform_int_distribution`: https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution

src/util.cc
```diff
@@ -24,6 +24,49 @@
...
+  static bool run = false;
+  if (!run) {
+    std::srand(std::time(nullptr));
+    run = true;
+  }
+}
+
+// get the port the local google sign in server will run on
+int getPort() {
+  static int port = 0;
```
nicholasbishop: I generally view hidden state like this to be an antipattern. If we need to cache the port, I'd prefer we do it outside the function.

src/oauth_server.h
```diff
@@ -0,0 +1,38 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef SRC_OAUTH_SERVER_H_
+#define SRC_OAUTH_SERVER_H_
+
+#include <QObject>
+#include <QString>
+
+#include <microhttpd.h>
```
nicholasbishop: If this is just for `struct MHD_Daemon`, we can just use a forward declaration and remove the include from the header file.

src/oauth_server.cc
```diff
@@ -0,0 +1,168 @@
...
+  if (daemon == NULL) {
+    daemon = MHD_start_daemon(MHD_USE_SELECT_INTERNALLY, gondar::getPort(),
+                              NULL, NULL, &answer_to_connection, this,
+                              MHD_OPTION_NOTIFY_COMPLETED, request_completed,
+                              NULL, MHD_OPTION_END);
+  }
+}
+
+void OauthServer::stop() {
+  if (daemon != NULL) {
```
nicholasbishop: Here and elsewhere, `nullptr` instead of `NULL`

src/oauth_server.cc
```diff
@@ -0,0 +1,168 @@
...
+OauthServer::OauthServer(QObject* parent) : QObject(parent) {
+  daemon = NULL;
+}
+
+void OauthServer::start() {
+  if (daemon == NULL) {
+    daemon = MHD_start_daemon(MHD_USE_SELECT_INTERNALLY, gondar::getPort(),
+                              NULL, NULL, &answer_to_connection, this,
+                              MHD_OPTION_NOTIFY_COMPLETED, request_completed,
+                              NULL, MHD_OPTION_END);
```
nicholasbishop: Recommend using local variables to give names to these nulls

src/oauth_server.h
```diff
@@ -0,0 +1,38 @@
...
+ public:
+  explicit OauthServer(QObject* parent = 0);
+  void start();
+  void stop();
+ signals:
+  void callbackReceived(QString state, QString code);
+  void authError(QString error);
+
+ private:
+  struct MHD_Daemon* daemon;
```
nicholasbishop: `struct MHD_Daemon* daemon = nullptr`;

src/oauth_server.cc
```diff
@@ -0,0 +1,168 @@
...
+    }
+  } else {
+    // a non-get request
+    // TODO(kendall): make separate pages for each error
+    return send_page(connection, errorpage);
+  }
+}
+
+OauthServer::OauthServer(QObject* parent) : QObject(parent) {
+  daemon = NULL;
```
nicholasbishop: This can be initialized in the header

src/oauth_server.cc
```diff
@@ -0,0 +1,168 @@
...
+
+static int answer_to_connection(void* cls,
+                                struct MHD_Connection* connection,
+                                const char* url,
+                                const char* method,
+                                const char* version,
+                                const char* upload_data,
+                                size_t* upload_data_size,
+                                void** con_cls) {
+  (void)version;          /* Unused. Silent compiler warning. */
```
nicholasbishop: Silent -> silence

kendase3: i should probably cite that this code is from gnu's included examples with libmicrohttpd

kendase3: see https://www.gnu.org/software/libmicrohttpd/tutorial.html#Hello-browser-example

.gitignore
kendase3: I've been running the various python scripts with my system python, which is still python2 on sid

kendase3: I just did a little more checking and it looks like it's python2 from basically every context.  it's a trivial amount of python though so it would be easy to switch to 3.  i don't think that's appropriate for this PR.

kendase3:
Just looked into the build failure; looks like I may have to fork libmicrohttpd:
fatal: unable to access 'https://gnunet.org/git/libmicrohttpd/': server certificate verification failed. CAfile: /etc/ssl/certs/ca-certificates.crt CRLfile: none
fatal: clone of 'https://gnunet.org/git/libmicrohttpd' into submodule path '/codebuild/output/src110821808/src/github.com/neverware/gondar/libmicrohttpd' failed
Failed to clone 'libmicrohttpd' a second time, aborting

For my own use cases I've been using ssh://...

kendase3:
While using our own fork resolved the certificate issue, I'll have to investigate why codebuild got a certificate error for GNU's source tree while I did not on my local machine.  I appear to have been using https for both cases.

kendase3:
😌 passed the new linter

justinoue:
Seems fine to me if it works (can't really comment on the details of the c++), my only input is that I thought we had determined to use a button image and not a text link?

Do you have a screenshot of what it looks like?

kendase3:
image button change is coming in a future PR.  it's text for this version.

kendase3:
there are some other goodies as well that would probably be quick to add (minor obfuscation of client secret) but this PR is already rather large.  this is basically the MVP version.

kendase3:
re specifics of what it looks like
i was just going to image capture the meepo one ^_^

justinoue:
sgtm

package.py
```diff
@@ -19,17 +18,27 @@ def run_cmd(*args):
...
-      cmd += ('--build-arg', 'CHROMEOVER=true',)
+        cmd += (
+            '--build-arg',
+            'CHROMEOVER=true',
+        )
+    if apikey:
+        cmd += ('--build-arg', 'METRICS_API_KEY={}'.format(apikey))
+    if googleclient:
+        cmd += ('--build-arg', 'GOOGLE_SIGN_IN_CLIENT={}'.format(googleclient))
     if apikey:
```
nicholasbishop: Is this `if` correct?

kendase3: good catch

kendase3:
please hold

kendase3:
I believe I addressed the feedback above

src/chromeover_login_page.cc
```diff
@@ -15,11 +15,38 @@
...
 #include "gondarsite.h"
 #include "gondarwizard.h"
 #include "log.h"
+#include "oauth_server.h"
 #include "util.h"
 
+// TODO(kendall): right now we are just using current time as rand seed
+// note: relies on initRand() having been called on startup
+static QString get_random_string() {
+  // most basic version first: all lowercase characters
```
nicholasbishop: Out of date comment?

src/meepo.cc
```diff
@@ -146,6 +163,16 @@ void Meepo::requestAuth(const QAuthenticator& auth) {
   network_manager_.post(request, doc.toJson(QJsonDocument::Compact));
 }
 
+void Meepo::requestGoogleAuth(QString id_token) {
+  QJsonObject json;
+  json["id_token"] = id_token;
+  LOG_WARNING << "id_token =" << id_token;
```
nicholasbishop: Why a warning?

src/meepo.h
```diff
@@ -37,6 +38,7 @@ class Meepo : public QObject {
   // sites. Call this method to initiate that flow, and await the
   // finished() signal to get the results (including errors).
   void start(const QAuthenticator& auth);
+  void startGoogle(QString id_token);
```
nicholasbishop: `const QString&`

src/chromeover_login_page.cc
```diff
@@ -39,22 +66,107 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
...
 }
 
+void ChromeoverLoginPage::handleGoogleSigninError(QString error) {
+  LOG_WARNING << "received google sign in error=" << error;
+}
+
+// This fires when the user has clicked on the 'sign in with google' link
+// we generate the URL to open in their browser, send them there, and start
+// the local server to parse the sign-in response
+void ChromeoverLoginPage::handleGoogleSigninPart1() {
```
nicholasbishop: Might be helpful to split these methods out into a new class, `GoogleSigninFlow` or similar. What do you think?

kendase3: There was a moment when I thought about where this logic should live organizationally.  It seemed reasonable either as part of the login page, as part of the meepo class, or as its own thing.  I figured I would just wait and see which you preferred.  the 'part1', 'part2'...were supposed to be stub names but I ended up liking them for making the sequence of events more clear.

src/meepo.cc
kendase3: i switched it to info.  i'm not sure how much of this if even appropriate to display as debug information.  the debug console is not present in a prod build, though.

src/chromeover_login_page.cc
nicholasbishop: I think I it's own class sounds good. Agree that the part1, part2 etc is good to make the sequence clear.

kendase3:

src/googleflow.cc
```diff
@@ -0,0 +1,121 @@
...
+
+void GoogleFlow::handleGoogleSigninError(QString error) {
+  LOG_WARNING << "received google sign in error=" << error;
+}
+
+// This fires when the user has clicked on the 'sign in with google' link
+// we generate the URL to open in their browser, send them there, and start
+// the local server to parse the sign-in response
+void GoogleFlow::handleGoogleSigninPart1() {
+  static bool http_server_started = false;
```
nicholasbishop: Can this be a field in the class instead?

PR #304, opened 2019-02-14
add lint target to linux dockerfile

kendase3:
This should enable linting on PRs

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,6 @@
 [submodule "gdisk"]
 	path = gdisk
 	url = https://github.com/neverware/gdisk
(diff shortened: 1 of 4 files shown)


kendase3:
according to https://answers.ros.org/question/285439/how-to-make-ros-cpplint-use-a-relative-path-for-header-guard-check/ it looks for the root of a git repository as one of its heuristics which would explain what i have observed.  i may copy a stub .git folder to the docker image if it solves the problem, as i'd like to get the header check back.

kendase3:
deprecated by https://github.com/neverware/gondar/pull/305

PR #306, opened 2019-02-21
Add xor layer for ssm values

kendase3:
Now split client secret in twain.  I also added some TODOs, most notably it looks like I'll have to go back to my stateful rand getter implementation in some form; only the first fetch appears to be random as-is.  There are a few things I'm still cleaning up with this but this is the general approach I'm taking.  We need python3.5 for tohex/fromhex functionality.


nicholasbishop:
I wonder if this isn't a bit overkill? I think slightly obfuscating secrets so that they don't show up with `strings` is reasonable, but maybe just use a static value for the XOR?

kendase3:
sure; that wouldn't change anything outside of the python script though, right?

kendase3:
or you're saying the c++ would have that value hardcoded?

kendase3:
after some meatspace discussion, closing this for now

PR #308, opened 2019-02-25
Fix random number generation on Windows

kendase3:
Fix random number generation on Windows  
This change uses different implementations of generating random bytes on Windows and on Linux. 

Addresses OVER-8937

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,7 +124,7 @@ add_library(app STATIC
   src/metric.cc
   src/neverware_unzipper.cc
   src/oauth_server.cc
-  src/randar.cc
+  src/rand_util.cc
   src/site_select_page.cc
   src/unzipthread.cc
   src/update_check.cc
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,7 @@ add_library(app STATIC
   src/metric.cc
   src/neverware_unzipper.cc
   src/oauth_server.cc
+  src/randar.cc
   src/site_select_page.cc
   src/unzipthread.cc
   src/update_check.cc
@@ -162,8 +163,7 @@ if(WIN32)
 
   fix_qt_static_link(app)
 
-  target_link_libraries(app PRIVATE setupapi)
-  target_link_libraries(app PRIVATE gdisk)
+  target_link_libraries(app PRIVATE setupapi gdisk bcrypt)
   # disable shadow variable checking for this file as it imports gdisk headers
   # which contain a shadowing whoopsie
   set_source_files_properties(src/gpt_pal.cc PROPERTIES COMPILE_FLAGS -Wno-shadow)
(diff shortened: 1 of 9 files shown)


kendase3:
self-assigning for now to address build problems

kendase3:
there was a bit of a gotcha with the include ordering that has been resolved.

src/randar.h
```diff
@@ -0,0 +1,47 @@
...
+
+namespace gondar {
+
+class RandomManager {
+ public:
+  static RandomManager& getInstance() {
+    static RandomManager instance;
+    return instance;
+  }
+  unsigned int getRandomByte();
```
nicholasbishop: Either the name or the return type seems wrong here

CMakeLists.txt
```diff
@@ -124,6 +124,7 @@ add_library(app STATIC
   src/metric.cc
   src/neverware_unzipper.cc
   src/oauth_server.cc
+  src/randar.cc
```
nicholasbishop: How about `random.cc`, `rand_util.cc`, or something like that?

src/main.cc
```diff
@@ -24,6 +24,7 @@
 #include "gondarwizard.h"
 #include "log.h"
 #include "metric.h"
+#include "randar.h"
```
nicholasbishop: Is this include used?

src/randar.cc
```diff
@@ -0,0 +1,68 @@
...
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#include "randar.h"
+
+#include <random>
+
+#include "log.h"
```
nicholasbishop: Needed?

src/randar.cc
```diff
@@ -0,0 +1,68 @@
...
+#if defined(Q_OS_WIN)
+unsigned int RandomManager::getRandom4Bytes() {
+  uint8_t Buffer[4];
+  uint16_t BufferSize;
+
+  BufferSize = sizeof(Buffer);
+  memset(Buffer, 0, BufferSize);
+
+  // fills buffer with random bytes
+  BCryptGenRandom(NULL, Buffer, BufferSize,
```
nicholasbishop: I assume this came from stackoverflow or similar, could you add a comment with a link for future reference?

src/randar.h
```diff
@@ -0,0 +1,47 @@
...
+// along with this program.  If not, see <http://www.gnu.org/licenses/>.
+
+#ifndef SRC_RANDAR_H_
+#define SRC_RANDAR_H_
+
+#include <random>
+
+namespace gondar {
+
+class RandomManager {
```
nicholasbishop: Is this class used outside of randar.cc? If not it doesn't need to be in the header.

src/randar.cc
kendase3: yeah came from a microsoft github example iirc; i'll find

kendase3:
@nicholasbishop take another gander when you get a chance

josephsussman:
A gander at [gondar](http://github.com/neverware/gondar).

PR #310, opened 2019-04-12
Handle error case for sign in with google

kendase3:
also remove unused func startServer

Addresses OVER-9058 and OVER-9059

in general pertains to OVER-7586

### DIFF
--- a/src/chromeover_login_page.cc
+++ b/src/chromeover_login_page.cc
@@ -25,6 +25,7 @@
 #include <cstdlib>
 #include <ctime>
 
+#include "googleflow.h"
 #include "gondarsite.h"
 #include "gondarwizard.h"
 #include "log.h"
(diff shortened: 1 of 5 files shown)


kendase3:
I tested this on the Linux client; haven't tested it on Windows yet

src/chromeover_login_page.cc
```diff
@@ -131,12 +133,22 @@ void ChromeoverLoginPage::handleMeepoFinished() {
 
 // This fires when we have received our token. We send the token to Meepo
 // to find the user's sites in a way very similar to user/password flow.
-// TODO(kendall): handle failure case
 void ChromeoverLoginPage::handleGoogleSigninFinished(QNetworkReply* reply) {
   LOG_INFO << "id token received; contacting meepo...";
+  // remove red error text that may have existed from a previous auth failure
+  meanWordsLabel.setVisible(false);
```
kendase3: this change is of dubious value; usually the meepo session will fire much faster than a user could observe

PR #311, opened 2019-04-15
google sign in uses button/image rather than link

kendase3:
see https://neverware.atlassian.net/browse/OVER-8860


src/chromeover_login_page.h
```diff
@@ -50,10 +51,11 @@ class ChromeoverLoginPage : public gondar::WizardPage {
   QLabel passwordLineEditLabel;
   QLabel meanWordsLabel;
   QLabel forgotLabel;
-  QLabel googleLabel;
+  QPushButton googleButton;
   bool finished;
   bool started;
   GoogleFlow googleFlow;
+  QPixmap googlepix;
```
nicholasbishop: `QPixmap` is [reference counted](https://doc.qt.io/qt-5/implicit-sharing.html), so you don't need to store it as a member of the class.

src/chromeover_login_page.cc
```diff
@@ -50,26 +50,32 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
...
-  googleLabel.setTextFormat(Qt::RichText);
-  googleLabel.setTextInteractionFlags(Qt::TextBrowserInteraction);
+
+  // image sourced from
+  // https://developers.google.com/identity/branding-guidelines
+  googlepix.load(":/images/btn_google_signin_light_normal_web.png");
+  googleButton.setIcon(googlepix);
+  googleButton.setIconSize(
+      QSize(googlepix.rect().width(), googlepix.rect().height()));
+  googleButton.setMaximumWidth(googlepix.rect().width());
```
nicholasbishop: I think we should probably remove the regular styling from the button, something like this:

```diff
diff --git a/resources/style.css b/resources/style.css
index c9b1edd..52f26e9 100644
--- a/resources/style.css
+++ b/resources/style.css
@@ -22,3 +22,7 @@ ImageSelectPage QLabel {
 * {
   font-size: 17px;
 }
+
+#googleSigninButton {
+    border: none;
+}
diff --git a/src/chromeover_login_page.cc b/src/chromeover_login_page.cc
index 138136b..cbe6a5e 100644
--- a/src/chromeover_login_page.cc
+++ b/src/chromeover_login_page.cc
@@ -58,6 +58,7 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
   googleButton.setIconSize(
       QSize(googlepix.rect().width(), googlepix.rect().height()));
   googleButton.setMaximumWidth(googlepix.rect().width());
+  googleButton.setObjectName("googleSigninButton");
 
   meanWordsLabel.setObjectName("loginError");
   layout.addWidget(&passwordLineEditLabel, 1, 0);
```

src/chromeover_login_page.cc
```diff
@@ -50,26 +50,32 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
   forgotLabel.setTextFormat(Qt::RichText);
   forgotLabel.setTextInteractionFlags(Qt::TextBrowserInteraction);
   forgotLabel.setOpenExternalLinks(true);
-  googleLabel.setText("<a href='#'>Sign in with Google</a>");
-  googleLabel.setTextFormat(Qt::RichText);
-  googleLabel.setTextInteractionFlags(Qt::TextBrowserInteraction);
+
+  // image sourced from
+  // https://developers.google.com/identity/branding-guidelines
+  googlepix.load(":/images/btn_google_signin_light_normal_web.png");
```
nicholasbishop: Interacting with the button feels a little weird because there are no hover or pressed states. I'd like to either add that to this PR, or file a ticket to do so in the future

kendase3: i can add

kendase3:
works good not great like this: it doesn't seem like there's an easy way to have a custom image for onclick, and the hover range is too wide

kendase3:
assigning myself for now; the current version has a weird hoverbox and the on-click image doesn't work yet

kendase3:

src/chromeover_login_page.cc
```diff
@@ -51,33 +51,57 @@ ChromeoverLoginPage::ChromeoverLoginPage(QWidget* parent) : WizardPage(parent) {
...
+}
+
+void ChromeoverLoginPage::showPressedButton() {
+  LOG_WARNING << "ken: button pressed~~~~";
+  setStyleSheet(
+      "#googleSigninButton { background-image: "
+      "url(':/images/btn_google_signin_light_pressed_web.png') }");
+}
+void ChromeoverLoginPage::showUnpressedButton() {
+  LOG_WARNING << "ken: button unpressed~~~~";
```
nicholasbishop: debug logs

PR #314, opened 2019-04-17
Fix previous implementation of success/fail messages

kendase3:
Addresses OVER-9059 by splitting signal/slots into success/fail

### DIFF
--- a/src/meepo.h
+++ b/src/meepo.h
@@ -46,6 +46,7 @@ class Meepo : public QObject {
 
  signals:
   void finished();
+  void failed(bool google_mode);
 
  private:
   void requestAuth(const QAuthenticator& auth);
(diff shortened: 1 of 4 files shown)


kendase3:
Hiyeran tested my previous implementation and it didn't display the expected string for when the user successfully signed into a google account but that account had no meepo account.

kendase3:
I tested this version with that case and it worked on my machine.

kendase3:
self-assigning for a second.  feel free to take a look but i'm looking over the new logic specifically to see if there are error flow regressions (the no-sites case and the meepo internal fail() calls that progress to error screen)

kendase3:
(raises paw)

src/meepo.h
```diff
@@ -72,6 +73,7 @@ class Meepo : public QObject {
   Sites sites_;
   QString error_;
   int sites_remaining_ = 0;
+  bool google_mode = false;
```
nicholasbishop: Should suffix with an underscore like the other fields

nicholasbishop: Also wouldn't hurt to add a comment here saying what google_mode is

PR #317, opened 2019-06-19
Use classic style instead of modern

kendase3:
This produces padding for the wizard's pixmap

Addresses OVER-9895

I'll have to double-check on Windows tomorrow but on Linux this resolves the issue.


kendase3:
tested this out; only marginally larger but avoids resizing issue (on my test machine at least) that justin observed

PR #320, opened 2020-02-19
Now link users to try.neverware.com when zero sites found

kendase3:
Addresses OVER-5156

### DIFF
--- a/src/error_page.cc
+++ b/src/error_page.cc
@@ -25,7 +25,13 @@ ErrorPage::ErrorPage(QWidget* parent) : WizardPage(parent) {
 }
 
 void ErrorPage::setErrorString(const QString& errorStringIn) {
+  // we can set the label so it may display links; it's ok
+  // if we don't use the link functionality in all cases
   label.setText(errorStringIn);
+  label.setTextFormat(Qt::RichText);
+  label.setTextInteractionFlags(Qt::TextBrowserInteraction);
+  label.setOpenExternalLinks(true);
+  label.setWordWrap(true);
 }
 
 bool ErrorPage::errorEmpty() const {
--- a/src/error_page.cc
+++ b/src/error_page.cc
@@ -21,6 +21,8 @@ ErrorPage::ErrorPage(QWidget* parent) : WizardPage(parent) {
   setSubTitle(" ");
   layout.addWidget(&label);
   label.setText("");
+  // we can set the label so it may display links; it's ok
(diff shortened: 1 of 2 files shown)


src/meepo.h
```diff
@@ -42,7 +42,9 @@ class Meepo : public QObject {
 
   QString error() const;
   Sites sites() const;
-  const QString no_sites_error = "User has no sites";
+  const QString no_sites_error =
+      "User has no sites.\nPlease visit <a "
```
nicholasbishop: The `\n` doesn't actually work. A `<br/>` would probably work, but I think having it all on one line is fine.

src/error_page.cc
```diff
@@ -25,7 +25,13 @@ ErrorPage::ErrorPage(QWidget* parent) : WizardPage(parent) {
 }
 
 void ErrorPage::setErrorString(const QString& errorStringIn) {
+  // we can set the label so it may display links; it's ok
+  // if we don't use the link functionality in all cases
   label.setText(errorStringIn);
+  label.setTextFormat(Qt::RichText);
+  label.setTextInteractionFlags(Qt::TextBrowserInteraction);
```
nicholasbishop: I don't think `setTextInteractionFlags` is needed

src/error_page.cc
```diff
@@ -25,7 +25,13 @@ ErrorPage::ErrorPage(QWidget* parent) : WizardPage(parent) {
...
+  label.setOpenExternalLinks(true);
+  label.setWordWrap(true);
```
nicholasbishop: These constant changes can be moved to the `ErrorPage` constructor rather than set each time `setErrorString` is called.

src/error_page.cc
nicholasbishop: Turns out that while this isn't needed, it is useful. It sets these flags: `TextSelectableByMouse | LinksAccessibleByMouse | LinksAccessibleByKeyboard`

src/error_page.cc
```diff
@@ -21,10 +21,16 @@ ErrorPage::ErrorPage(QWidget* parent) : WizardPage(parent) {
...
   setLayout(&layout);
 }
 
 void ErrorPage::setErrorString(const QString& errorStringIn) {
+  // we can set the label so it may display links; it's ok
+  // if we don't use the link functionality in all cases
```
nicholasbishop: Comment seems like it belongs in the constructor

PR #321, opened 2020-02-21
Beerover Gondar should no longer supply 32-bit images: Addresses OVER-11238

kendase3:
Refactors newestImage to belong to the wizard as a whole rather than an individual page now that beerover users are skipping image selection page in its entirety.

Addresses OVER-11238

### DIFF
--- a/src/gondarwizard.h
+++ b/src/gondarwizard.h
@@ -26,6 +26,7 @@
 #include "device_picker.h"
 #include "download_progress_page.h"
 #include "image_select_page.h"
+#include "newest_image_url.h"
 #include "usb_insert_page.h"
 #include "write_operation_page.h"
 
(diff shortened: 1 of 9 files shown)


kendase3:
I haven't tested this yet and am tidying it up still; but this is the refactor I have in mind.

src/gondarwizard.h
```diff
@@ -59,11 +60,17 @@ class GondarWizard : public QWizard {
   DownloadProgressPage downloadProgressPage;
   UsbInsertPage usbInsertPage;
   WriteOperationPage writeOperationPage;
+  NewestImageUrl newestImageUrl;
 
   const std::vector<GondarSite>& sites() const;
   void setSites(const std::vector<GondarSite>& sites);
   bool isFormatOnly() const { return formatOnly; }
   void setFormatOnly(bool newValue) { formatOnly = newValue; }
+  bool getError();
```
nicholasbishop: What's this new global error state needed for?

kendase3: previously error states were page by page, but the latest image url fetching now happens on different pages for chromeover/beerover, so made sense to move it to the wizard itself.  in general these error state bools exist so validatepage can allow passing through to the 'next' page for error cases (the next page being the error page for these)

kendase3:
the reason for this is that there's only one image for beerover right now; so seems silly to have a large % of our users go through a flow where they hit a page with one button and have to hit next.  previously this logic was handled on the image select page.  i agree it feels like there's no really great satisfying home for this; in the past it was also kind of weird that one page would go back and ask the image select page for the right url.  i'd be fine with reorganizing it if you can think of something less confusing.

kendase3:
the error bit could be moved to the individual page for example (at least for now, it's conceivable other pages would require a validate page short-circuit like this to use next() on error), it would just make the transaction a little more complicated

nicholasbishop:
Yeah, I completely agree that the beerover image selection page should be removed. I think is fine as is for now.

PR #324, opened 2020-02-26
update column limit

kendase3:


.clang-format
```diff
@@ -1,2 +1,3 @@
 BasedOnStyle: Chromium
 SortIncludes: true
+ColumnLimit: 80
```
nicholasbishop: When I run `clang-format --style=Chromium --dump-config` it shows the `ColumnLimit` is already 80. I don't mind this being explicit though.

kendase3: for me without this had a behavioral change (though not in current master)

PR #325, opened 2020-02-26
Gondar feedback: OVER-9424

kendase3:
Initial feedback functionality for gondar.  User will supply site if they have progressed far enough into the wizard to have a site associated.  This version uses local debug port as the server; this will be updated in a subsequent PR once the server-side changes are complete.

Addresses OVER-9424

### DIFF
--- a/src/feedback_dialog.cc
+++ b/src/feedback_dialog.cc
@@ -49,7 +49,8 @@ FeedbackDialog::FeedbackDialog() {
 
 static QNetworkRequest createFeedbackRequest() {
   // FIXME(ken): send to the real endpoint once it exists
-  auto url = QUrl("https://udrhaj2mil.execute-api.us-east-1.amazonaws.com/dev/gondar");
+  auto url =
+      QUrl("https://udrhaj2mil.execute-api.us-east-1.amazonaws.com/dev/gondar");
   QNetworkRequest request(url);
   request.setHeader(QNetworkRequest::ContentTypeHeader, "application/json");
   return request;
@@ -94,8 +95,7 @@ void FeedbackDialog::submit() {
     json["product"] = "beerover";
   }
   QJsonDocument doc(json);
-  LOG_INFO << "feedback: " << std::endl
-              << doc.toJson(QJsonDocument::Indented);
+  LOG_INFO << "feedback: " << std::endl << doc.toJson(QJsonDocument::Indented);
   network_manager_.post(request, doc.toJson(QJsonDocument::Compact));
 
   // clear text field in case user wants to send more feedback later
--- a/src/feedback_dialog.cc
+++ b/src/feedback_dialog.cc
@@ -34,6 +34,8 @@ FeedbackDialog::FeedbackDialog() {
           &FeedbackDialog::submit);
   connect(&network_manager_, &QNetworkAccessManager::finished, this,
           &FeedbackDialog::handleReply);
+  connect(&feedback_field_, &QTextEdit::textChanged, this,
+          &FeedbackDialog::maybeEnableSubmit);
 
   layout_.addWidget(&feedback_label_);
   layout_.addWidget(&feedback_field_);
@@ -41,6 +43,8 @@ FeedbackDialog::FeedbackDialog() {
   setLayout(&layout_);
   setMinimumWidth(500);
   setWindowTitle(tr("CloudReady USB Maker"));
+  // submit button initially grayed out, there's no input yet
+  submit_button_.setEnabled(false);
 }
 
 static QNetworkRequest createFeedbackRequest() {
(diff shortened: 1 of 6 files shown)


src/feedback_dialog.h
```diff
@@ -0,0 +1,50 @@
...
+
+ public:
+  FeedbackDialog();
+  void submit();
+
+ private:
+  void handleReply(QNetworkReply* reply);
+  QLabel feedback_label_;
+  QLabel title_label_;
+  QLineEdit title_;
```
nicholasbishop: I'd suggest not having a separate title field. We can always derive a summary by taking the first line of the details field, if that's useful.

This matches the cloudready feedback dialog:
image

src/feedback_dialog.cc
```diff
@@ -0,0 +1,91 @@
...
+// the user's UUID
+// any site information (do we want to only offer this after sign-in phase?)
+void FeedbackDialog::submit() {
+  LOG_WARNING << "the feedback was accepted";
+  auto request = createFeedbackRequest();
+  QJsonObject json;
+  json["title"] = title_.text();
+  json["details"] = details_.toPlainText();
+  json["uuid"] = gondar::GetUuid();
+  json["site"] = gondar::GetSiteId();
```
nicholasbishop: Should probably have a field to indicate beerover vs chromeover, maybe `product`

src/feedback_dialog.cc
```diff
@@ -0,0 +1,91 @@
...
+    LOG_INFO << "Successfully sent feedback";
+  }
+}
+
+// we collect:
+// the title of the ticket
+// the ticket contents
+// the user's UUID
+// any site information (do we want to only offer this after sign-in phase?)
+void FeedbackDialog::submit() {
```
nicholasbishop: Currently the user is able to submit even with nothing in the text box. I think it makes sense to gray out the Submit button until there's something in the text box, or at least skip sending feedback in that case.

src/feedback_dialog.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef SRC_FEEDBACK_DIALOG_H_
+#define SRC_FEEDBACK_DIALOG_H_
+
+#include <QDialog>
+#include <QLabel>
+#include <QLineEdit>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
+#include <QPushButton>
+#include <QTextBrowser>
```
nicholasbishop: QTextBrowser -> QTextExit

src/feedback_dialog.cc
```diff
@@ -0,0 +1,104 @@
...
+    LOG_ERROR << "Error response from feedback server";
+  } else {
+    LOG_INFO << "Successfully sent feedback";
+  }
+}
+
+// if the feedback field is non-empty, enable submit button
+void FeedbackDialog::maybeEnableSubmit() {
+  QString content = feedback_field_.toPlainText();
+  if (content.length() == 0) {
```
nicholasbishop: `content.isEmpty()`

src/feedback_dialog.cc
```diff
@@ -0,0 +1,104 @@
...
+  } else {
+    json["product"] = "beerover";
+  }
+  QJsonDocument doc(json);
+  LOG_WARNING << "feedback: " << std::endl
+              << doc.toJson(QJsonDocument::Indented);
+  network_manager_.post(request, doc.toJson(QJsonDocument::Compact));
+
+  // closes the window
+  accept();
```
nicholasbishop: Might want to also clear the text area, otherwise if the user opens the feedback dialog again after submitting the previous text will still be there.

src/feedback_dialog.cc
```diff
@@ -0,0 +1,104 @@
...
+  json["feedback"] = feedback_field_.toPlainText();
+  json["uuid"] = gondar::GetUuid();
+  json["site"] = gondar::GetSiteId();
+  if (gondar::isChromeover()) {
+    json["product"] = "chromeover";
+  } else {
+    json["product"] = "beerover";
+  }
+  QJsonDocument doc(json);
+  LOG_WARNING << "feedback: " << std::endl
```
nicholasbishop: `LOG_WARNING` -> `LOG_INFO`

src/feedback_dialog.cc
```diff
@@ -0,0 +1,104 @@
...
+    submit_button_.setEnabled(true);
+  }
+}
+
+// we collect:
+// the ticket contents
+// the user's UUID
+// any site information (do we want to only offer this after sign-in phase?)
+void FeedbackDialog::submit() {
+  LOG_WARNING << "the feedback was submitted";
```
nicholasbishop: Same here, `LOG_INFO`

PR #327, opened 2020-02-27
on crostini this downloads folder does not exist

kendase3:


src/downloader.cc
```diff
@@ -62,6 +62,10 @@ void DownloadManager::startNextDownload() {
 
   const QDir dir =
       QStandardPaths::writableLocation(QStandardPaths::DownloadLocation);
+  if (!dir.exists()) {
+    // equivalent of mkdir -p
+    dir.mkpath(".");
```
nicholasbishop: Worth checking the return value here? It could at least log a failure

src/downloader.cc
```diff
@@ -62,6 +62,14 @@ void DownloadManager::startNextDownload() {
 
   const QDir dir =
       QStandardPaths::writableLocation(QStandardPaths::DownloadLocation);
+  if (!dir.exists()) {
+    // equivalent of mkdir -p
+    bool success = dir.mkpath(".");
+    if (!success) {
+      LOG_ERROR << "Could not create download directory: "
+                << QStandardPaths::DownloadLocation;
```
nicholasbishop: I think you want to output `dir` here, `QStandardPaths::DownloadLocation` is just the integer 14.

kendase3: :axe_smart:

PR #328, opened 2020-03-18
Now paginate site results for users with a great many sites: OVER-11068

kendase3:
Addresses OVER-11068

### DIFF
--- a/src/meepo.cc
+++ b/src/meepo.cc
@@ -116,6 +116,7 @@ std::vector<GondarSite> sitesFromReply(const QJsonArray& rawSites) {
   return sites;
 }
 
+// Get the next page from the pagination dict. If on the last page, return 0.
 int getNextPage(const QJsonObject& outer_json) {
   const QJsonObject json = outer_json["pagination"].toObject();
   auto cur = json.value("current").toInt();
--- a/src/meepo.cc
+++ b/src/meepo.cc
@@ -209,7 +209,6 @@ void Meepo::handleSitesReply(QNetworkReply* reply) {
   const QJsonObject json = gondar::jsonFromReply(reply);
   const QJsonArray rawSites = json["sites"].toArray();
   auto new_sites = sitesFromReply(rawSites);
-  // FIXME(ken): this insert causing a leak
   sites_.insert(sites_.end(), new_sites.begin(), new_sites.end());
 
   LOG_INFO << "received " << sites_.size() << " site(s)";
--- a/src/meepo.cc
+++ b/src/meepo.cc
@@ -82,10 +82,11 @@ QNetworkRequest createGoogleAuthRequest() {
   return request;
 }
 
-QNetworkRequest createSitesRequest(const QString& api_token) {
+QNetworkRequest createSitesRequest(const QString& api_token, int page) {
   auto url = createUrl(path_sites);
   QUrlQuery query;
   query.addQueryItem("token", api_token);
+  query.addQueryItem("page", QString::number(page));
   url.setQuery(query);
   return QNetworkRequest(url);
 }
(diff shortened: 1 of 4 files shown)


src/meepo.cc
```diff
@@ -116,6 +116,17 @@ std::vector<GondarSite> sitesFromReply(QNetworkReply* reply) {
   return sites;
 }
 
+int getNextPage(const QJsonObject& outer_json) {
```
nicholasbishop: Docstring here would be helpful. Something like this, assuming I understood it correctly:
```
// Get the next page from the pagination dict. If already on the last page, return 0.
```

src/site_select_page.h
```diff
@@ -35,8 +37,18 @@ class SiteSelectPage : public gondar::WizardPage {
   bool validatePage() override;
 
  private:
+  int getTotalPages();
+  void handleNextPage();
+  void handlePrevPage();
+  void updateSitesForPage();
   QButtonGroup sitesButtons;
   QVBoxLayout layout;
+  QGroupBox pageGroup;
```
nicholasbishop: I think it might be a little nicer to put the page label inline with the buttons, something like this (without the red border of course):
image

nicholasbishop:
If I'm reading this right it prefetches all the sites at the beginning, so the pagination aspect of the API isn't really used? If so, might make sense to just use a QListWidget (https://doc.qt.io/qt-5/qlistwidget.html#details) or similar rather than constructing pages of radio buttons.

kendase3:
that's correct; it prefetches all images for all sites.  i don't think it's much data.  we could certainly use a qlistwidget or any other abstraction that might be easier for users to pick from.  i believe one suggestion was a search field?  i have no preference and we could try out a few pretty quickly and see what we like.

src/meepo.cc
kendase3: yeah, that's accurate

nicholasbishop:
Ah yeah a search widget sounds good to me. That would pair well with the list widget.

kendase3:
going to look into this search business next, but @nicholasbishop take a look and see if this is what you had in mind for the list

nicholasbishop:
Yep, the new widget looks good to me. Is it in a state now where you want me to do more review?

kendase3:
you can if you'd like; i'm just planning on filtering visible items based on a qtextedit field for search.  everything else is pretty much just as i'd envisaged.

kendase3:
@nicholasbishop take another look when you get a chance; not sure about "find" and "search" text vs. other words and not sure about appropriateness of public data members, but this is what i had in mind instead of the pagination

nicholasbishop:
Minor UI suggestion: have the search widget filter the list as the user types rather than an explicit Find button.

kendase3:
sure, can remove the find button

nicholasbishop:
Looking good. A minor thing I noticed is that if I select a site, then type stuff into the search field such that the selected site is no longer visible, I can still click Next to advance to the next page. Not a big deal but a little surprising. I think keeping the selection as the user types into the search field is good though, so maybe just add a final check to see if the selected site is visible?

src/site_select_page.h
```diff
@@ -16,27 +16,43 @@
...
+  QHBoxLayout layout;
+};
+
 class SiteSelectPage : public gondar::WizardPage {
   Q_OBJECT
 
  public:
   explicit SiteSelectPage(QWidget* parent = 0);
+  QString getFindText();
+  FindWidget find;
```
nicholasbishop: This doesn't need to be public

src/site_select_page.h
```diff
@@ -16,27 +16,43 @@
...
+  Q_OBJECT
+ public:
+  explicit FindWidget(QWidget* parent = nullptr);
+  QString getFindText();
+  QLineEdit lineEdit;
+
+ private:
+  QLabel findLabel;
+  QHBoxLayout layout;
+};
```
nicholasbishop: I don't think this needs to be a separate widget, you can just add the hbox, line edit, and label to the SiteSelectPage.

kendase3: yeah i just divided it for organizational clarity.  i can consolidate

src/site_select_page.h
```diff
@@ -29,14 +30,19 @@ class SiteSelectPage : public gondar::WizardPage {
...
+  QString getFindText();
 
  protected:
   void initializePage() override;
   bool validatePage() override;
 
  private:
-  QButtonGroup sitesButtons;
+  QLabel findLabel;
+  QHBoxLayout findLayout;
```
nicholasbishop: The hbox should come before its children to fix a crash on exit

kendase3: i think i naively just :sorted them

PR #329, opened 2020-04-01
Now send metrics to meepo as well when we have a valid session token: OVER-8821

kendase3:
Note that this PR includes a rework to move some newly-shared functionality up to the wizard-level out of individual pages.

Currently the "description" field includes version and value fields, as meepo appears to be able to infer everything else from the token such as site.

Addresses OVER-8821

### DIFF
--- a/src/gondarwizard.cc
+++ b/src/gondarwizard.cc
@@ -19,6 +19,7 @@
 
 #include "about_dialog.h"
 #include "admin_check_page.h"
+#include "chromeover_login_page.h"
 #include "device_select_page.h"
 #include "error_page.h"
 #include "feedback_dialog.h"
@@ -38,6 +39,7 @@ class GondarWizard::Private {
 
   AdminCheckPage adminCheckPage;
   DeviceSelectPage deviceSelectPage;
+  ChromeoverLoginPage chromeoverLoginPage;
   SiteSelectPage siteSelectPage;
   ErrorPage errorPage;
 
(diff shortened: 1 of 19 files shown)


kendase3:
i just remembered that @justinoue asked me to preface these strings with usbmaker; i'll take care of that in the morn'.

kendase3:
update: prefaced with usb-maker as requested

src/meepo.cc
```diff
@@ -281,13 +287,63 @@ void Meepo::handleDownloadsReply(QNetworkReply* reply) {
...
+  // but a day may come when we add stuff like this to beerover, so
+  // there's no harm in adding a guard here
+  if (isChromeover() && siteId != 0) {
+    inner_json.insert("site_id", siteId);
+  }
+  json["activity"] = inner_json;
+  QNetworkRequest request(url);
+  request.setHeader(QNetworkRequest::ContentTypeHeader, "application/json");
+  QJsonDocument doc(json);
+  QString strJson(doc.toJson(QJsonDocument::Compact));
```
nicholasbishop, line 327: Might be worth splitting the stuff above where it sets up the request into a separate function so that you can add a test in https://github.com/neverware/gondar/blob/master/test/test.cc

src/gondarwizard.cc
```diff
@@ -39,7 +38,6 @@ class GondarWizard::Private {
 
   AdminCheckPage adminCheckPage;
   DeviceSelectPage deviceSelectPage;
-  ChromeoverLoginPage chromeoverLoginPage;
```
nicholasbishop, line 42: Why did this need to move?

kendase3: good point; i've since removed the logic that required this change

src/meepo.cc
kendase3: sure i can do that

kendase3:
made a small change to add the tests as you requested @nicholasbishop

nicholasbishop:
tests failed though

kendase3:
the tests failed because i had not considered how that test would run in beerover mode; hopefully fixed

PR #336, opened 2020-09-24
update gondar-release to call python3

kendase3:


kendase3:
looks like at least the print statements will be ok :monocle_face:

nicholasbishop (approved):

